PR #7, opened 2017-02-04
Module touchscreen

Lastrellik:
First off, I am new to this. Please feel free to give me any all feedback. This is my understanding of OOP best practice and makes sense to me, but this is just based on what I've read and learned in the past. 

This is the framework for the Touchscreen module. I added the package in the src folder and updated the gitignore from gitignore.io. From there I created a few classes for the application, the GUI and the front end. My thought is that the application is just the main class where everything will be centralized. I think as we (potentially) add more classes we can have them all centralized there. The TouchscreenGUI.java will handle the front end of the code and TouchscreenBackEnd.java will handle functionality. TouchscreenGUI.java was created from Windowbuilder in Eclipse to help make the GUI there.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,85 @@
 .piolibdeps
 .clang_complete
 .gcc-flags.json
+/bin/
+
+
+
+# Created by https://www.gitignore.io/api/code-java,java,eclipse
+
+### Code-Java ###
+# Language Support for Java(TM) by Red Hat extension for Visual Studio Code - https://marketplace.visualstudio.com/items?itemName=redhat.java
+
+.project
+.classpath
+factoryConfiguration.json
+
+### Eclipse ###
+
+.metadata
+bin/
+tmp/
+*.tmp
+*.bak
+*.swp
+*~.nib
+local.properties
+.settings/
+.loadpath
+.recommenders
+
+# Eclipse Core
+
+# External tool builders
+.externalToolBuilders/
+
+# Locally stored "Eclipse launch configurations"
+*.launch
+
+# PyDev specific (Python IDE for Eclipse)
+*.pydevproject
+
+# CDT-specific (C/C++ Development Tooling)
+.cproject
+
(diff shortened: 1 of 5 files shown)


src/module_TOUCHSCREEN/TouchscreenApp.java
```diff
@@ -0,0 +1,21 @@
+/**
+ * Salt Lake Community College
+ * Spring 2017 Semester
+ * CSIS-2810 Computer Architecture
+ * Instructor: Jon McGowan
```
Zeustopher, line 5: I have hesitations about this particular header on each class file. Were this an "assignment" then I wouldn't mind it but in a project like this I'd probably suggest we remove it or put in a header specific to this project and have our names on it. Lets discuss this on Monday. I won't have this stop the pull request.

.gitignore
```diff
@@ -2,3 +2,85 @@
...
+
+# Package Files #
+*.jar
+*.war
+*.ear
+
+# virtual machine crash logs, see http://www.java.com/en/download/help/error_hotspot.xml
+hs_err_pid*
+
+# End of https://www.gitignore.io/api/code-java,java,eclipse
```
Zeustopher, line 86: Probably a little overkill on the .gitignore, but hey - at least now we've got all our bases covered 😛

src/module_TOUCHSCREEN/TouchscreenApp.java
```diff
@@ -0,0 +1,21 @@
...
+ * 
+ */
+
+package module_TOUCHSCREEN;
```
Zeustopher, line 9: I've not ever dealt with an Arduino or the compile/upload process that it goes through. Its possible having both Java and C++ code inside the same source directory might cause issues, but I really have no idea. If it works then this will be a good structure. If it doesn't then we can figure out a structure that will and isolate each language from the other. We can worry about that when we get to it.

Zeustopher (approved):
I made a few comments, but other than that I'd say it looks good.

PR #8, opened 2017-02-04
Added Base Wiring Diagrams

Zeustopher:
I've created the Base Module Wiring Diagram. 

I intentionally tried to keep everything on one half of the board as much as possible so that we can use the other half for future modules. I tried to also limit the pins being used to the pins in one section so that we have a large area to be able to connect modules to.

(_This was based off the wiring diagram found here: http://keithkay.com/2013/03/03/arduino-self-calibrating-laser-trip-wire/_)


Lastrellik:
Looks good man. I have no idea if this wiring diagram would actually work, but it looks cool! I'm sure we'll find out.

PR #10, opened 2017-02-10
Added "Mini" Module

Zeustopher:
I was thinking that it might be nice to provide a "Mini" module which operates on only the bare essentials of a single resistor, the photoresistor, and uses the built in LED. All the details aren't flushed out yet, but it was a good starting point.

I also added a table to the modules' readme. I think there might be some modules that won't play nicely together or are redundant. Thhis will give us a way of documenting that.


Lastrellik:
This is a great idea.

PR #12, opened 2017-02-16
basic flash

Boomstick105:
basic master code


Lastrellik:
So this branch was 19 commits behind the master. It's important to git pull before contributing or else it would break the repo. If you have time this weekend I can help you out a bit.

PR #14, opened 2017-02-18
Setup board

Lastrellik:
I went ahead and made it a little more object oriented and created a ComponentTester class. This class is an OO approach to testing components and I think should be the first thing that happens when you turn on the System. I also mapped all the pins in the main to the drawing @Zeustopher  created. My thought is that we could make main.cpp an object as well, but we have to have the "setup" and "loop" functions outside of that class, so I don't know if it really makes sense to put it in a class. I'm open to suggestions.

I also commented out the "TestSerialConnection.cpp" because it didn't like that there were two classes with "setup" and "loop" inside the project.

### DIFF
--- a/src/ComponentTester.cpp
+++ b/src/ComponentTester.cpp
@@ -1,0 +1,64 @@
+#include "Arduino.h"
+class ComponentTester {
+
+private:
+  int testedPin; //the pin to test
+  int millisOn = 500;
+  int millisOff = 500;
+  String pinMode;
+
+public:
+  //for testing output
+  ComponentTester(int pin, String mode) {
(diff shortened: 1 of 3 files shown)


Zeustopher (approved):
I had a similar idea. I actually was going to make each component a class and then pass that component into a tester similar to this.

I'm in the middle of doing my classes so I'll push those up in a little while and then we can decide how best to try and adapt them into this tester class.

Also - the C++ convention for class files is to have both a *.h (header) and a *.cpp (class) file which backs it. I was planning on having my classes follow that convention. Lets decide in my pull request whether we want to do it that way or whether we want to include the *.cpp directly.

PR #15, opened 2017-02-18
Added Component Classes

Zeustopher:
Added the following classes to the project:
- LED
- PhotoResistor
- Buzzer

Changed main.cpp to use the new classes.

Note for future refactor potential: 
- _Instead of passing in the pin and the string to the tester we could pass in the component itself then have the tester pull out the type from the object passed in and do the log that way._ 
- _There is duplicated code in these classes for getting pin, componentType, etc. This suggests we should make a super-class for these classes. Possibly call it "Component". Then refactor these classes to make them subclasses of that class._

### DIFF
--- a/src/Buzzer.cpp
+++ b/src/Buzzer.cpp
@@ -1,0 +1,23 @@
+/*
+  Photoresistor.h - Simple Library for reading from a Photoresistor
+*/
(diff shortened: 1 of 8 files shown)


Lastrellik:
Nice. I love it.

PR #19, opened 2017-02-19
Component super class

Lastrellik:
I created the Component superclass with the respective getters and setters for the pin and component type. I also went through each of the component classes and had them all extend Component and removed the unnecessary code.

Please keep in mind that although it compiles, I'm not entirely sure that it will work just yet. My next task is to modify the ComponentTester class to not take a String as an argument, but just the component itself. Then I'll see if each of the components still work. I'll try to get that done tonight.

### DIFF
--- a/src/Button.cpp
+++ b/src/Button.cpp
@@ -4,19 +4,13 @@
 #include "Arduino.h"
 #include "Button.h"
 
-Button::Button(int pin) {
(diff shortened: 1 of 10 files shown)


Zeustopher (approved):

Zeustopher:
I haven't tested this on my own board yet, but everything looks in order.

PR #20, opened 2017-02-19
Fix component tester

Lastrellik:
I went through and refactored each of the components to extend Component and removed the duplicated code. I also refactored ComponentTester so that it is able to extract the ComponentType from the object passed to it.

I had to spend some time learning the concept of passing by reference and passing by value in C++. This process had me modifying a lot of code throughout the project, but hopefully I didn't break anything :)

Each of the element tests are working now. I think we can start moving on to actual alarm functionality and the like with this framework.

### DIFF
--- a/src/Button.cpp
+++ b/src/Button.cpp
@@ -4,19 +4,13 @@
 #include "Arduino.h"
 #include "Button.h"
 
-Button::Button(int pin) {
+Button::Button(int pin) : Component(pin, "BUTTON"){
   pinMode(pin, INPUT);
-  _pin = pin;
-  _componentType = "BUTTON";
-}
-
-int Button::getPin() {
-  return _pin;
 }
 
 boolean Button::isPressed() {
   boolean pressed = false;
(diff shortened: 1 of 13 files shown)


Zeustopher:
I'm not sure how the Git repository is going to take it when we merge in this Pull Request. This pull request is from a branch off another branch and actually contains some of the same commits I just merged in with the previous pull request... Github isn't complaining about there being any conflicts. I'm just not sure how the git history is going to look after this one (and it looks like the next one is in the same boat). Let me go do some research and figure out what we can do to mitigate any issues that might come up. Maybe there is a way that we can 'rebase' this Pull Request with the latest changes in master or something...

Boomstick105:
Whatever works, works for me, just keep me informed

Zeustopher:
Oopsey... I just somehow pushed this into Master directly from my local box... I guess I better just merge this one in then...

Git sure does let you have enough rope to hang yourself with...

PR #21, opened 2017-02-20
Laser module

Lastrellik:
I created a laser module class and added it to the main class and component tester class. I also went and added the laser module to the wiring diagram.


Zeustopher:
This was a little hard to sift through because it contained changes from the last two Pull Requests as well. I'm really not sure how to avoid these types of "cascading" Pull Requests in the future. Lets brainstorm on Wednesday after/before class on a strategy for the future.

There *must* be a way to 'rebase' a Pull Requests from Master but I'm just not sure how to do what I want to have happen... I'll draw on a whiteboard or something on Wednesday.

Overall - great work on these last few pull requests. (I haven't tested it on my board yet but I'm assuming it works well)

PR #26, opened 2017-02-25
Build alarm hardcode method

Lastrellik:
I know it can't automatically merge. It's late.

I decided to scrap the idea of the Builder method. After doing much studying on references and pointers and everything I decided that the Builder method was more for building complex objects where you had to take certain steps at a time in order to create the object (Like if you were making a Pizza object you first have to turn on the oven, make the dough, spread the sauce, etc). Once it was done we would have had to hard code the back end anyway. 

I'm open to suggestions of either scrapping this code or accepting it and modifying it. I'm also open to the idea if any of you think there's a better way to do this. 

Anyway, it's 1:43 AM. I'm going to bed. I basically created the alarm object and got it mostly working. There's still some work to be done, but we're almost there.


Zeustopher:
There was a conflict that I had to resolve related to my previous refactor of the ComponentTester.

Zeustopher (approved):
I'm in favor of this change.

The situation it sounded like you ran into I think was probably the one that was nagging in the back of my brain when we were talking about it after class on Wednesday about just needing to end up hardcoding it all anyway.

Since I don't have a laser unit controlled by the Arduino (_maybe I should buy one... @sithboy: Do you have one of these laser? Maybe we should buy another set and go in halfs for it._) it was hard for me to test all the functionality. Since mine doesn't finish calibrating (I need to figure out exactly when I need to shine the laser on the sensor) once both LEDs turn off I can't get them to turn back on again even after I press the button. I hear the sound when I press the button but nothing happens with the LEDs. BUT, what I can test seems good. We can continue to make tweaks, etc.

Overall, I like the way it is looking so far. I'm giving my approval. I'm not going to merge it in because I want to get the others take on it before we do.

sithboy:
I don't have a wired laser but I'll order one right now.

PR #33, opened 2017-03-02
Clean Code and improve Base

Lastrellik:
So as I mentioned in class, this is the refactor I did after reading a few chapters in that book.

I also was able to add a bit of functionality and make it more of a usable alarm system. I think with the base where it's at we're ready to start working on different modules. 

It's starting to look pretty sweet!

### DIFF
--- a/src/LED.cpp
+++ b/src/LED.cpp
@@ -15,12 +15,27 @@ LED::LED() : Component(13, "LED"){
 }
 
 void LED::on() {
-  digitalWrite(Component::getPin(), HIGH);
+  digitalWrite(this->getPin(), HIGH);
   _isOn = true;
 }
 
 void LED::off() {
-  digitalWrite(Component::getPin(), LOW);
+  digitalWrite(this->getPin(), LOW);
+  _isOn = false;
+}
+
+void LED::flash(){
+  this->on();
+  delay(50);
+  this->off();
+  delay(30);
+  _isOn = false;
+}
+
+void LED::flash(int millisOn){
+  this->on();
+  delay(millisOn);
+  this->off();
   _isOn = false;
 }
 
--- a/src/Laser.cpp
+++ b/src/Laser.cpp
@@ -11,12 +11,12 @@ Laser::Laser(int pin) : Component(pin, "LASER"){
 Laser::Laser(){
 }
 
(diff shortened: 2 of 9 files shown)


src/LED.cpp
```diff
@@ -15,12 +15,27 @@ LED::LED() : Component(13, "LED"){
...
+  digitalWrite(this->getPin(), LOW);
+  _isOn = false;
+}
+
+void LED::flash(){
+  this->on();
+  delay(50);
+  this->off();
+  delay(30);
+  _isOn = false;
```
Zeustopher, line 32: This is redundant since we are already setting _isOn to false inside the off() method. We can leave it here though - better safe than sorry.

src/LED.cpp
```diff
@@ -15,12 +15,27 @@ LED::LED() : Component(13, "LED"){
...
+}
+
+void LED::flash(int millisOn){
+  this->on();
+  delay(millisOn);
+  this->off();
```
Zeustopher, line 38: In the other flash() method there is also a delay after turning the LED off. Is there a reason there isn't one here / there is on there?

src/Laser.cpp
```diff
@@ -11,12 +11,12 @@ Laser::Laser(int pin) : Component(pin, "LASER"){
 Laser::Laser(){
 }
 
-void Laser::engage(){
+void Laser::on(){
```
Zeustopher, line 14: "engage" was such a cool word for the laser though ;-)

src/LED.cpp
Lastrellik: My thought process at the time was that I was going to use it to communicate a message to the user, like "Task Failed" or something like that. Then it would just move onto the next task. I may have just missed the delay afterwards. I'm down for whatever.

src/Laser.cpp
Lastrellik: Haha one of the points the book made was to not be cute. For example, someone may put a funny inside joke method signature inside the code not expecting it to be a source of confusion somewhere down the road. That section of the book made me specifically think about this.

src/LED.cpp
Zeustopher: Lets just leave it as is for now. We can address it later if we need to.

PR #36, opened 2017-03-08
Updated Dev Environment Setup Instructions

Zeustopher:
We decided that the Git Atom Plugin was 'faulty' and didn't want it included in the setup instructions any longer. We also found that Clang wasn't installed on several other platforms so I took the Windows Only message out of the header.


sithboy:
This look ok to me, Chris. thanks for the heads up.

PR #37, opened 2017-03-09
Wifi Module - Initial Work

Zeustopher:
Note that there is still LOTS of work to be done for the Wifi Module... This is simply a convenient place to merge this into master so that we can start taking advantage of the new Properties class throughout the rest of the project.

When reviewing this Pull Request it might be easier if you go through it commit by commit. This way you can see my thinking as I moved through each step.


sithboy (approved):
Looks good for a rough draft of this.

sithboy (commented):
Looks good.

PR #41, opened 2017-03-14
Add debug logging back in

Zeustopher:
We only want to send a bunch of debug logging statements if it is configured in Properties.cpp to be turned on. This change adds checks before each print statement to see whether it is enabled and only log if it is.

It also adds a reference to the Properties.h file to those classes which didn't have it yet (even if they aren't currently debug logging anything).

I reverted (most of) the original change which took these out and then added the checks mentioned above.

Note that **Debug Logging** is now  **disabled by default**. The assumption here is that in a 'production' install, it is assumed that you don't need that constant stream of data happening and would only want to turn that on manually when you need to troubleshoot a device. Having it turned on is the _exception_, not the _rule_.

### DIFF
--- a/src/ComponentTester.cpp
+++ b/src/ComponentTester.cpp
@@ -27,17 +27,19 @@ void ComponentTester::testPin(){
   }
   if(componentType.equalsIgnoreCase("PHOTORESISTOR")){
     for(int i = 1; i <= 10; i++){
-        /*Serial.print("test#");
+        Serial.print("test#");
         Serial.print(i);
         Serial.print(" ");
-        Serial.println(analogRead(pin));*/
+        Serial.println(analogRead(pin));
         delay(100);
       }
   }
   if(componentType.equalsIgnoreCase("BUTTON")){
     Button button(_component.getPin()); //May be wasted memory
+    Serial.println("Press the button to test");
     for(int i = 0; i < 150; i++){
       if(button.isPressed()) {
+        Serial.println("The button was pressed");
         break;
       }
       delay(200);
--- a/src/ComponentTester.cpp
+++ b/src/ComponentTester.cpp
@@ -28,7 +28,7 @@ void ComponentTester::testPin(){
   if(componentType.equalsIgnoreCase("PHOTORESISTOR")){
     for(int i = 1; i <= 10; i++){
         if(Properties::DEBUGGING_ACTIVE) {
-          Serial.print("test# " + i + ": ");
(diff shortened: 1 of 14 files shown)


src/ComponentTester.cpp
```diff
@@ -27,17 +27,19 @@ void ComponentTester::testPin(){
   }
   if(componentType.equalsIgnoreCase("PHOTORESISTOR")){
     for(int i = 1; i <= 10; i++){
-        /*Serial.print("test#");
-        Serial.print(i);
-        Serial.print(" ");
-        Serial.println(analogRead(pin));*/
+        if(Properties::DEBUGGING_ACTIVE) {
+          Serial.print("test# " + i + ": ");
```
Lastrellik: So this line of code actually breaks the program. I don't know why C++ does this, but you can't easily concatenate Strings natively in c++. They need to be separated into multiple Serial.print() commands.

src/main.cpp
```diff
@@ -53,10 +53,12 @@ void setup() {
 void loop(){
   if(not alarm->isArmed()){
     if(alarm->isButtonPressed()){
+      SerialComm::sendLogMessage("Alarm Armed");
```
Lastrellik: So with how the Raspberry PI works, this line of code will significantly slows down the arming process because it keeps waiting for a response from the PI after sending this line of code. This should only run if Properties::Module_PI is set to true.

src/main.cpp
```diff
@@ -53,10 +53,12 @@ void setup() {
...
       alarm->arm();
     }
   } else {
     if(alarm->isTripped()){
+      SerialComm::sendLogMessage("Alarm Tripped");
```
Lastrellik: Same as above. SerialComm::sendLogMessage() sends a packet and waits for the SHA1 hash response. Since nothing is listening when it sends the packet out, no response comes back and it times out.

Lastrellik (changes requested):
This is good stuff! A few of the debugging commands are a little out of place, but I'm all for keeping this in for debugging purposes.

src/ComponentTester.cpp
Zeustopher: Ah - ok. Good catch. I didn't have my environment all configured on my work laptop so I didn't try to compile. I'll fix this.

src/main.cpp
Zeustopher: I'll add that.

Zeustopher:
@Lastrellik - Ready for a re-review. Corrected all issues you pointed out.

PR #43, opened 2017-03-15
Initial Turret Module Work

sithboy:
This commit is @sithboy's that got misplaced onto an old branch.

Contains an initial wiring diagram, updated readme file, and some initial code work.


Lastrellik:
@sithboy nice work on this! I was talking to some engineers at one of my client's and they got all excited about this idea and how it could be done. I encourage you to keep at it and let me know if you would like any help. Maybe we could schedule a time to work on it together or if you get stuck and have any questions.

docs/circuit_diagrams/module_TURRET/VertandHoriz.ino
```diff
@@ -0,0 +1,53 @@
+#include <Servo.h>
+
+Servo vert,hori;  // create servo object to control a servo
+
+int pos = 0;    // variable to store the servo position
+int laser = 12;
+int buzzer = 11;
```
Lastrellik, line 7: We can use existing code to use objects for these, even though this is ultimately what's going on under the hood.

docs/circuit_diagrams/module_TURRET/VertandHoriz.ino
```diff
@@ -0,0 +1,53 @@
...
+     }
+     delay(timeBetweenBursts);
+}
+
+void fire(){
+     digitalWrite(laser,HIGH);
+     analogWrite(buzzer,100);
+     delay(20);//adjust this to change length of turret shot
+     digitalWrite(laser,LOW);
+     analogWrite(buzzer, 0);
```
Lastrellik, line 52: Like stated above, these can be represented by existing code and objects. Another thing I learned in that Clean Code book is to instead of commenting how this code works, maybe create a variable called "lengthOfTurretShot" and let the code speak for itself.

Lastrellik (approved):
I'm excited to see where this goes! I'll approve this but I think it'd be a good idea to make those changes I mentioned in here sometime in one of the next commits. Keep it up, man.

PR #44, opened 2017-03-19
Fix bug and improve alarm

Lastrellik:
Most of the work here is through troubleshooting the alarm command listener for the raspberry pi. After HOURS of debugging I think I found the root cause of my sea of frustration in the Alarm::trigger() function.

This pull request fixes that bug and allows for disarm capability if the alarm is armed.


src/ComponentTester.cpp
```diff
@@ -26,20 +31,17 @@ void ComponentTester::testPin(){
     tone(pin, 1000, 100); //(Buzzer)_component.tone(1000) wouldn't work. Polymorphism?
   }
   if(componentType.equalsIgnoreCase("PHOTORESISTOR")){
-    for(int i = 1; i <= 10; i++){
+    /*for(int i = 1; i <= 10; i++){
```
Zeustopher: If we are just commenting this out, what purpose does it serve to have a test for the photoresistor?

src/Logger.h
```diff
@@ -0,0 +1,21 @@
+/*
+Logger.h - A class based around logging events
```
Zeustopher, line 2: This class was a great idea. 👍

src/Properties.cpp
```diff
@@ -12,7 +12,7 @@
 #include "Properties.h"
 
 // ********** GENERAL ********** //
-boolean Properties::DEBUGGING_ACTIVE = false;
+boolean Properties::DEBUGGING_ACTIVE = true;
```
Zeustopher: Based on my last commit when I flipped this to false, I think we want to keep it that way. We want to assume that we don't want the noise and only turn it on when we do. Lets set this back to false.

src/main.cpp
```diff
@@ -37,10 +39,12 @@ void testBoardComponents(){
 }
 
 void setup() {
-  Serial.begin(Properties::BAUD_RATE);  //Begin serial communication
-  testBoardComponents();
+  delay(5000);
+  Serial.begin(Properties::BAUD_RATE);
+  //testBoardComponents();
```
Zeustopher: Was turning this off in main intentional? If there are circumstances where we don't want testing to take place then we should wrap this with an if-check and put another entry in the Properties class with a name something like "TEST_COMPONENTS" or something.

src/Component.cpp
```diff
@@ -3,12 +3,12 @@
...
 #include "Arduino.h"
 #include "Component.h"
+#include "Logger.h"
 
 Component::Component(int pin, String componentType){
-  if(Properties::DEBUGGING_ACTIVE) Serial.println("Component " + componentType + " is set at pin " + pin);
+  Logger::Log("Component " + componentType + " is set at pin " + pin);
   _pin = pin;
   _componentType = componentType;
-  delay(100);
```
Zeustopher: This was the "fix" we implemented before for that 'race condition' we were having. Did another fix take care of that so this small delay is no longer needed? What fix was that?

Zeustopher (changes requested):

Zeustopher:
I wasn't sure where best to put this as a comment on the code so I'll just put it out here. I noticed alot of code related to communicating with the PI itself but this pull request seemed to be initially related to just fixing a bug and adding logging. It also looks like this branch and the `PI_Commands_To_Arduino are at the exact same point in the commit tree.

Was it intentional to include those other serial communication / PI communication changes on this Pull Request?

_(If it was, I need to go back and take a look at those changes. I skipped them until I saw the answer to the above question)_

src/ComponentTester.cpp
Lastrellik: So I'm having this error where incorrect parameters are being passed to my logger class.... The weirdest things break the code. I'll have to tell you about it tomorrow. This was one that was breaking the code randomly that I commented out but didn't uncomment. We can fix that.

Lastrellik:
@Zeustopher So, I had this experience where I was debugging this WEIRD stuff that keeps happening and somewhere along the lines I broke the code and couldn't get it to fix. My last commit was forever before that and it was bad enough and I was frustrated enough that I just restored the code from a local backup that fixed the issue. The problem is that I had a lot of random changes with throughout the whole project that I didn't want to lose. Outside of the PI module everything works flawlessly but inside the module it's being very shady... Like the messages I'm sending to the logger class cause a weird shift in memory and all my messages get screwed up. I'd really like to team up and get your take on what the heck is happening because I'm getting really stumped.

This commit was just an attempt to put another checkpoint in what I had coded so far. I think I need to learn to slow down and focus on one problem at a time and stop going all over the place with the code. I'm still pretty new to this. Any and all feedback is appreciated.

Lastrellik:
@Zeustopher I ran the fixes you suggested. To give you an idea of what's happening, can you try turning debugging mode on and uncomment the line of code in under Alarm::trigger()? Just to try to rule out faulty hardware on this Arduino.... I can't believe how difficult this has been. 

For me the arduino prints a blank line instead of that string. Nothing in this world makes sense anymore.

Zeustopher:
We're going to go in another direction instead of this one. Closing this Pull Request.

PR #45, opened 2017-03-23
Add Logging Documentation

Zeustopher:
We decided that we need to change the way we are logging these. We discussed as a group a Logging & Command Message scheme we could start to use. This commit adds the documenation around that scheme we came up with.

At this point, none of this documentation has been implemented. We wanted the documenation in place before we started the refactor.

Note - the values currently in the `Details` column of the table are just placeholders and will be updated with actual values as we perform the refactor.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,7 @@ Development
 ===========
 
 * [IDE / Dev Environment Setup Instructions](./docs/development/dev_environment_setup.md)
+* [Logging & Command Message Details](./docs/development/logging.md)
 
 
 Wiring Diagrams / Schematics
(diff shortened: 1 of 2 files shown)


Zeustopher:
Helpful hint for the reviewers: If you click the `View` button at the top of the file in the Files changed tab you can preview it with all of the Markdown formatting which is easier to read through than just the plaintext version.

Lastrellik (approved):
Yeah man looks good!

Lastrellik:
I think we can make a CSV that our tool works directly with that will also be used by the PI. I can't wait to start working with Java again :)

PR #46, opened 2017-03-23
Nuke and renew

Lastrellik:
I removed all SerialComm and DataPacket code. I still want to have a CommandListener, but it won't be based on DataPackets anymore, but just simple bytes. This commit also includes the improvements I've made to the alarm system.


Zeustopher:
I will admit that I didn't look at this Pull Request in a huge amount of detail - there are alot of changes to sift through. Since we're mainly wanting this one to go in to get us in a 'fresh state' then I'm just going to merge it as-is. If there are any issues in it we will just address it moving forward.

@Lastrellik, @sithboy, @Boomstick105 - Lets all make sure that we have rebased to main so we are all on the exact same page once this goes in. That way we can identify any issues that may have cropped up because of this change.

PR #48, opened 2017-03-24
Create modified command listener on uno

Lastrellik:
This is SOOOOOO much easier since I no longer have to deal with SHA1 or string BS or any other convoluted C++ madness.

This pull request contains the code from my previous pull request plus code that allows the Uno to respond to commands from the Raspberry PI. Right now the Raspberry PI code is set up to start the Arduino and then arm it after a few seconds and then disarm it. This is done by sending a single byte through the Serial Port to be deciphered by the Arduino's CommandListener.

### DIFF
--- a/src/Alarm.cpp
+++ b/src/Alarm.cpp
@@ -27,6 +27,7 @@ void Alarm::calibrate(){
     Logger::Log("Alarm Successful Calibration");
     _isCalibrated = true;
   }
+  _laser->off();
 }
 
 void Alarm::determineBasePhotoresistorReading(){
@@ -55,6 +56,7 @@ void Alarm::alertSuccessfulAction(){
 }
 
 void Alarm::arm(){
+  _laser->on();
   if (!this->isReadyToArm()){
    Logger::Log("Alarm failed to arm");
    this->alertFailedAction();
@@ -82,8 +84,9 @@ bool Alarm::isTripped(){
 }
 
 void Alarm::trigger(){
-  //Logger::Log("Alarm has been triggered!");
+  Logger::Log("Alarm has been triggered!");
   _isTriggered = true;
+  _isSilenced = false;
 }
 
 void Alarm::soundOneAlarmCycle(){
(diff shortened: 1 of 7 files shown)


Zeustopher:
I updated this pull request to not show the previous Pull Request's changes as well. _(It was hard to sort through them and know what I've already looked at)_

Zeustopher:
@Lastrellik - I was under the impression based on Wednesday night's discussion that the same 'scheme' would be employed both on the PI side and Arduino side. Ithought the Arduino would be accepting the same 5-digit integer that it would be sending to the PI. So I would have expected a code something similar to `30001` to be passed into the Arduino where `3` represented a Command type message, `0` was NA for logging level, and if `001` was defined as being the Arm command.

Lastrellik:
@zeustopher that shouldn't be too hard. I mostly just need a table for what the three digits numbers mean. Did we have someone working on that? Or was it supposed to be me?

Zeustopher:
@Lastrellik - I know @Boomstick105 was going to be working on replacing the logging but I don't know that we assigned it out to anyone to do the commands. I'm in that file right now documenting a "Code Comment Convention" with these messages so I'll quickly define codes for your commands you have right now (arm, disarm, etc.). I'll get that pushed up in a few minutes and you can make the modifications to this PR to have the Arduino recognize them.

Lastrellik:
@zeustopher have you ran the Java side of things? You should be able to just plug the arduino in and run the Java code from Eclipse and it should automatically select the arduino and run it. I haven't tested it in Linux though.

Zeustopher:
@Lastrellik - PR https://github.com/Lastrellik/Calico-Home-Security/pull/49 is ready with the codes documented.

I haven't yet though I will this afternoon once I get back to working on it.

Lastrellik:
@Zeustopher Can you take a look at it now and tell me what you think?

Zeustopher:
@Lastrellik - Please update the table with those two additional commands for `Execute Command: Trigger` and `Execute Command: ResetCalibration`.

PR #50, opened 2017-03-25
Move Component Testing Code to Alarm Class

Lastrellik:
I moved the test components code over to the Alarm class to allow all alarm actions to be controlled via the Raspberry PI if that module is connected. It also eliminated the need to make more of the component objects in memory by just utilizing the already-created objects in the Alarm class. 

Now the alarm can run standalone or it can run as a slave to the Raspberry PI.

### DIFF
--- a/src/module_PI/Arduino_Uno/CommandListener.cpp
+++ b/src/module_PI/Arduino_Uno/CommandListener.cpp
@@ -54,5 +54,8 @@ void CommandListener::executeCommand(byte* commandBytes){
     case 30006:
       alarm->resetCalibration();
       break;
+    case 30007:
+      alarm->testBoardComponents();
+      break;
   }
 }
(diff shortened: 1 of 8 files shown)


src/module_PI/Arduino_Uno/CommandListener.cpp
```diff
@@ -54,5 +54,8 @@ void CommandListener::executeCommand(byte* commandBytes){
     case 30006:
       alarm->resetCalibration();
       break;
+    case 30007:
```
Zeustopher, line 57: Please add these commands to the [Logging Readme](https://github.com/Lastrellik/Calico-Home-Security/blob/master/docs/development/logging.md) so that they stay in sync. This applies to this one and the two others above that aren't in the readme either.

Zeustopher (changes requested):

Lastrellik:
@Zeustopher Thanks for letting me know. It is done.

Boomstick105 (approved):
Yeah man looks good now.

PR #51, opened 2017-03-25
Pull Request of Serial Write Commands

Boomstick105:


docs/development/logging.md
```diff
@@ -49,12 +49,20 @@ Message Translation Chart
...
+| | | 021 - Base Photoresistor Data |
+| | | 022 - Base Photoresistor Test |
+| | | 023 - Component Created |
+| | | 024 - Component Tester Created |
+| | | 025 - Button Pressed |
+| | | 026 - CommandListener Created |
 
 **Details Block Definitions:**
-* [001 - 009] - Reserved for Commands intended for the Alarm (Arduino) to execute.
+* [001 - 030] - Reserved for Commands intended for the Alarm (Arduino) to execute.
```
Zeustopher, line 65: These aren't commands that the Arduino will be executing that were added. These are just logging statements. This should be changed back to `[001-009]`. If you want to 'carve out' another block for these, thats fine though we probably don't need it for these though.

src/Alarm.cpp
```diff
@@ -14,7 +14,7 @@
 #include "ComponentTester.h"
 
 Alarm::Alarm(){
-  Logger::Log("Alarm has been created");
+  Serial.write(14016); //14016 = Log, Info Alarm Created
```
Zeustopher, line 17: This is a little nit-picky, but this doesn't quite match the comment convention defined in the [Developer Code Comment Convention Documentation](https://github.com/Lastrellik/Calico-Home-Security/blob/4f540c026c61163b1bf5dc57533bb2c4d0915a3d/docs/development/logging.md#developer-code-comment-convention). It should be `// ####### = X, Y, Z`. You've got it as `// ####### = X, Y Z` - note the missing comma between `Y` and `Z`.

There are several of these scattered throughout this Pull Request.

src/Alarm.cpp
```diff
@@ -53,31 +53,31 @@ void Alarm::determineBasePhotoresistorReading(){
     _greenLED->flash();
   }
   _baseReading = avgReading / numOfReadings;
-  Logger::Log("Base Photoresistor Reading Determined");
+  Serial.write(24021); //24021 = Data, Info, Base Photoresistor Data
```
Zeustopher, line 56: This isn't quite right. In the notes of the [Message Translation Chart](https://github.com/Lastrellik/Calico-Home-Security/blob/4f540c026c61163b1bf5dc57533bb2c4d0915a3d/docs/development/logging.md#message-translation-chart) it points out that `Log Level` is only relevant for Log Type (1) messages. Since this is a Data message (2) then the `Log Level` number should be `0` for `N/A`

src/Alarm.cpp
```diff
@@ -53,31 +53,31 @@ void Alarm::determineBasePhotoresistorReading(){
...
   _laser->on();
   if (!this->isReadyToArm()){
-   Logger::Log("Alarm failed to arm");
+   Serial.write(14019); //14011 = Log, Info, Alarm Failed Arming
    this->alertFailedAction();
    if(!this->isArmed()) _laser->off();
  } else {
    this->alertSuccessfulAction();
-    Logger::Log("Alarm sucessfully armed");
+    Serial.write(14018); //14018 = Log, Info, Alarm Successful
```
Zeustopher, line 80: I think that, based on the additions to the Translation Chart, that it would be more appropriate to send a `020`. I think that was the original intention but it was just overlooked.

src/Alarm.cpp
```diff
@@ -53,31 +53,31 @@ void Alarm::determineBasePhotoresistorReading(){
...
+  Serial.write(14018); //14018 = Log, Info, Alarm Successful
   _buzzer->soundAffirmativeTone();
   _greenLED->flash(1000);
 }
 
 void Alarm::arm(){
   _laser->on();
   if (!this->isReadyToArm()){
-   Logger::Log("Alarm failed to arm");
+   Serial.write(14019); //14011 = Log, Info, Alarm Failed Arming
```
Zeustopher, line 75: Your comment is wrong. It has `011` instead of `019`. This will be confusing for someone coming in later.

src/Alarm.cpp
```diff
@@ -99,7 +99,7 @@ bool Alarm::isTripped(){
 }
 
 void Alarm::trigger(){
-  Logger::Log("Alarm has been triggered!");
+  Serial.write(34005); //34005 = Command, Info, Excuet Command: Trigger
```
Zeustopher, line 102: Two things:
1. You've got a typo in `Excuet`.
2. Are you meaning to send a command to the Pi to have the Pi execute something? I would have thought this would have been just a logging statement instead of sending a command for the Pi to execute. Or if that was the intention then it should _also_ send a logging message.

src/Alarm.cpp
```diff
@@ -112,6 +112,7 @@ void Alarm::soundOneAlarmCycle(){
 }
 
 void Alarm::resetCalibration(){
+  Serial.write(14006); //14006 = Log, Info, Execute Command: Reset Calibration
```
Zeustopher, line 115: `006` isn't really appropriate here. Thats an execute command. It needs to have its own logging statement.

src/Alarm.cpp
```diff
@@ -127,11 +128,11 @@ void Alarm::disarm(){
   _alarmLED->off();
   _isSilenced = false;
   _laser->off();
-  Logger::Log("Alarm has been disarmed");
+  Serial.write(34002); //34002 = Command, Info, Execute Command: Disarm
```
Zeustopher, line 131: `002` isn't really appropriate here. Thats an execute command. It needs to have its own logging statement. It should say something about 'successfully disarmed' or something.

src/Alarm.cpp
```diff
@@ -127,11 +128,11 @@ void Alarm::disarm(){
...
 }
 void Alarm::silence(){
   _isSilenced = true;
-  Logger::Log("Alarm has been silenced");
+  Serial.write(34003); //34003 = Command, Info, Excuet Command: Silence
```
Zeustopher, line 135: Same as above. `003` isn't appropriate here.

src/Buzzer.cpp
```diff
@@ -12,7 +12,6 @@ Buzzer::Buzzer(){
 }
 
 void Buzzer::soundTone(int frequency) {
-  if(Properties::DEBUGGING_ACTIVE) Serial.println("Buzzer has started sounding");
```
Zeustopher, line 15: Was it intentional that these two logging statements were removed instead of just changed into the new format?

src/Component.cpp
```diff
@@ -6,7 +6,7 @@
 #include "Logger.h"
 
 Component::Component(int pin, String componentType){
-  Logger::Log("Component " + componentType + " is set at pin " + pin);
+  Serial.write(14023); //14023 = Log, Info, Componed Created
```
Zeustopher, line 9: Typo in `Componed`.

src/ComponentTester.cpp
```diff
@@ -9,11 +9,11 @@
 #include "Logger.h"
 
 ComponentTester::ComponentTester() {
-  Logger::Log("Default ComponentTester Constructor");
+  Serial.write(14024); //14024 = Log, Info, Componet Tester
```
Zeustopher, line 12: Shouldn't this be `Component Tester Created` instead of `Componet (typo) Tester`? Same thing for the one below.

src/ComponentTester.cpp
```diff
@@ -30,18 +30,16 @@ void ComponentTester::testPin(){
...
-      }
+        Serial.write(14022); //14022 Log, Info, Base Photoresistor Test
   }
+
   if(componentType.equalsIgnoreCase("BUTTON")){
     Button button(_component->getPin()); //May be wasted memory
     for(int i = 0; i < 150; i++){
       if(button.isPressed()) {
-        Logger::Log("The button was pressed");
+        Serial.write(24017); //24017 Data, Info, Button Pressed
```
Zeustopher, line 42: This should be a logging statement OR the `Log Level` should be `0`. Also - it should be `025` for `Button Pressed` - not `017` which is for `Alarm Failed`.

Zeustopher (changes requested):

Zeustopher:
@Boomstick105 - Sorry that some of these seem really nit-picky / pedantic but a few of them can have really detrimental side effects further down the road when another developer comes behind you and tries to 'translate' the messages.

src/Buzzer.cpp
Boomstick105: Not it wasn't intentional. I'm going to scrap the request for the pull entirely as it's not cleary up to par, and use your comments as notes, on redoing it. 

I'm also going to add a problem for documenting the code thoroughly so I'm able to understand it better and push it so if others look at it they can understand as well.

PR #54, opened 2017-03-26
Documentation of Main Code

Boomstick105:
This includes documentaion of the main code to make it easier to read to future devs or future debugging.

There may be some typos at time of submitting, will correct if found, or edited as other push and pulls are made.

### DIFF
--- a/src/Alarm.cpp
+++ b/src/Alarm.cpp
@@ -1,7 +1,15 @@
-/*
-  Alarm.h - A class for creating, arming, disarming, and calibrating the alarm
+/**
+  Calico Home Security System, Alarm.cpp
+  Purpose: Alarm creaties an alarm ojbect, arms it, disarms it, and calibrates the paramaters
+  for the alarm object
+
+  @author Chris Nash, Jason Bruderer, David Tille, Tyler Jacobs
+  @version To be Determined
 */
 
+/**
+
+*/
 #include "Arduino.h"
 #include "Component.h"
 #include "LED.h"
@@ -13,11 +21,20 @@
 #include "Logger.h"
 #include "ComponentTester.h"
 
+/**
+  Base contructor that writes that the Alarm Object has been created
+*/
 Alarm::Alarm(){
   Logger::Log("Alarm has been created");
 }
 
-
+/**
+  Takes in the base calibration of the light level using the
+    determineBasePhotoresistorReading function and alerts of failed action
+      if alertFailedAction and return isCalibrated as false, else it returns
+      isCalibrated as true.
+    @param isCalibrated is set to true or false
+*/
 void Alarm::calibrate(){
   determineBasePhotoresistorReading();
   if(_photoR->takeReading() - _threshold < _baseReading){
@@ -30,7 +47,10 @@ void Alarm::calibrate(){
   }
   _laser->off();
 }
-
+/**
+  Creates a ComponentTester object that makes sure that all components are
+    running
+*/
 void Alarm::testBoardComponents(){
   if (isArmed()) disarm();
   ComponentTester tester(_greenLED);
@@ -43,7 +63,14 @@ void Alarm::testBoardComponents(){
   //tester.testComponent(_armButton);
 
 }
-
+/**
+  Determines a base of the photorsistor at normal light levels and then
+    determines the value of the photorsistor when the laser is on it.
+  @param avgReading a base for photorsistor at normal light, taken over time,
+    determine the average light level of the room.
+  @param numOfReadings hardcoded number used a for loop to help determines the
+    base reading.
+*/
 void Alarm::determineBasePhotoresistorReading(){
   _laser->off();
   int avgReading = 0;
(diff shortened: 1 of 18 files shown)


src/Alarm.cpp
```diff
@@ -82,44 +120,71 @@ void Alarm::arm(){
...
 void Alarm::trigger(){
   Logger::Log("Alarm has been triggered!");
   _isTriggered = true;
   _isSilenced = false;
 }
-
+/**
+  If _isSilenced is not set soundAlarmHighTone and soundAlarmLowTone are used
+    to set tone high and low, also tells _redLED and _alarmLED to flash
+*/
```
Lastrellik, line 159: The reason for this method is that I ran into an issue in the main class where if the alarm was triggered it wouldn't listen to any other commands sent to it. If you notice in the loop now it still checks for things like "Is the button pressed" etc, but sounds this single alarm cycle every loop. This allows us to still send commands to the Arduino even when it's triggered.

src/Alarm.h
```diff
@@ -33,6 +38,20 @@ class Alarm{
...
+      @param _greenLED, _redLED, _alarmLED used for LEDs on the breadboard
+        creates LED objects to manage them
+      @param _photoR used for value of photorsistor
+        creates Photoresistor object to manage it
+      @param _buzzer used to set Buzzer value
+        creates Buzzer object to manage it
+      @param _armButton used to determine state of Button on or off
+        creates a Button object to manage it
+      @param _laser used to manage the laser status on or off
+        creats a Lser object to manage it
```
Lastrellik, line 51: Just a few typos in here :)

src/Buzzer.cpp
```diff
@@ -1,44 +1,83 @@
...
+    tone function.
+    This version of the soundTone Function does not use duration with the tone
+    function.
+*/
 void Buzzer::soundTone(int frequency) {
+  //Need to take a look at while we're using properies for a Serail.println
+  //does not affected the tone
   if(Properties::DEBUGGING_ACTIVE) Serial.println("Buzzer has started sounding");
   tone(this->getPin(), frequency);
 }
```
Lastrellik, line 32: I think this may have been one that I forgot to switch over to the Logger::Log paradigm. I don't know if we really need to log this anyway, but I guess we can. Maybe at the Trace log level.

src/LED.cpp
```diff
@@ -1,43 +1,69 @@
...
+*/
 void LED::flash(){
   this->on();
   delay(70);
   this->off();
   delay(50);
 }
-
+/**
+  Flashes the LED on and off by using fucntions of off() and on()
```
Lastrellik, line 54: Small typo here

src/Laser.cpp
```diff
@@ -1,26 +1,42 @@
-/*
-  Laser.cpp - Simmple library for the laser module
+/**
+  Calico Home Security System, Laser.h
+  Purpose: Manges the lASER and creats a LASER object
```
Lastrellik, line 3: Few typos here

Lastrellik (changes requested):
I couldn't find much that wasn't accurate! I made comments on the typos and reasoning behind stuff. Ultimately I think this will be very useful, I just think it's important that we remember to update the comments and the code whenever we make changes.

Boomstick105:
I'm need to start copying and pasting my code into word and checking for those typos, and thanks for looking over it. Reading over all this again and again my mind sometimes doesn't pick things up. 

And thanks

Zeustopher:
Yeah - Atom's spell checking is less than great...

@Boomstick105 (and @sithboy and @Lastrellik if you're also interested) - There is a way to configure Atom's Spell Checker plugin to also spell check inside c++ source code files.

To add the setting necessary, go to `File --> Settings`. Click on `Packages` in the side menu. In the searhc box type `spell`. You should see an entry for `spell-check`. Click on the `Settings` button. Under the `Settings` section there is a `Grammars` field. Click into that field to edit it and paste `, source.cpp` at the end of it. (So the entire line should be `source.asciidoc, source.gfm, text.git-commit, text.plain, text.plain.null-grammar, source.cpp`) Click out of that field to save the setting. Close the Settings tab. Open (or close the reopen) any `.cpp` or `.h` file and it will now be spell checking it.

_**BIG NOTE / GOTCHA**_ - This is going to add _ALOT_ of false positives on spelling errors though. This plugin isn't really designed to spellcheck within a C++ or other language file. As such, it is going to complain about basically everything in that file. So if you do add this configuration to the plugin, you're going to want to only enable spell checking when doing basically what you're working on now - working with comments / documentation. To toggle it on/off go to `Packages --> Spell Check --> Toggle`.

Zeustopher:
@Boomstick105 - Two questions on this one:

1. I am having build errors on this branch. They are related to there not being a Logger.h file. Is this branch building properly on your end?
2. Is the intention still for us to finish this Pull Request / Merge this branch? It seems like many of the changes here are also duplicated in https://github.com/Lastrellik/Calico-Home-Security/pull/55 so I'm confused on whether I should be merging this PR or not...

Boomstick105:
I built this because I didn't know how long it would take be to get the documentation and the table up and pushed it to see if the notes were good, and Jason checked over them.

We can get rid of this pull request because my code with the table already includes them.

Zeustopher:
Ok. Closing this PR and nuking the branch.

PR #55, opened 2017-03-27
Completed Serial.Write() Table

Boomstick105:
Creats a table for Serial.Write() commands

Notes;
For the Component.cpp I did this for now, if we need to figure out what out in the comment section, I'll work on this asap, but for now I got most of the working table up and running.

Checked it a couple times, hopefully this is in the correct format and typo free, if not I'll look into addressing what needs to be addressed.
 
Serial.write(13250); // 13250 = Log, Debug, Component object successfully created
//Need to figure out if we can put the component type created and the pin it's
  //set at with the Serial.write and the table, for now just doing base message
  //Logger::Log("Component " + componentType + " is set at pin " + pin);

### DIFF
--- a/docs/development/logging.md
+++ b/docs/development/logging.md
@@ -49,14 +49,45 @@ Message Translation Chart
 | 5 - Request | 5 - Trace        | 005 - Execute Command: Trigger |
 |             |                  | 006 - Execute Command: Reset Calibration |
 |             |                  | 007 - Execute Command: Test Components |
+|             |                  | 100 - Alarm object successfully created |
+|             |                  | 101 - Alarm base photoresistor reading determined
+|             |                  | 102 - Alarm failed action
+|             |                  | 103 - Alarm successful action
+|             |                  | 104 - Alarm failed to arm
+|             |                  | 105 - Alarm successfully armed
+|             |                  | 150 - Button object successfully created
+|             |                  | 151 - Button has been pressed
+|             |                  | 200 - Buzzer object has been created
+|             |                  | 201 - Buzzer has started to sound
+|             |                  | 202 - Buzzer has stopped sounding
+|             |                  | 250 - Component object successfully created
+|             |                  | 300 - ComponentTester default constructor object successfully created
+|             |                  | 301 - ComponentTester object successfully created
+|             |                  | 350 - Laser object successfully created
+|             |                  | 351 - Laser has been turned on
+|             |                  | 352 - Laser has been turned off
+|             |                  | 400 - LED object successfully created
+|             |                  | 401 - LED has been turned on
+|             |                  | 402 - LED has been turned off
+|             |                  | 403 - LED flashing
+|             |                  | 404 - LED toggling
+|             |                  | 450 - Photoresistor object has been successfully created
+|             |                  | 451 - Photoresistor taking reading
 |             |                  | 514 - [EXAMPLE] Component Configured on Pin 14 |
 |             |                  | 515 - [EXAMPLE] Component Configured on Pin 15 |
 |             |                  | 516 - [EXAMPLE] ... |
 
 **Details Block Definitions:**
 * [001 - 009] - Reserved for Commands intended for the Alarm (Arduino) to execute.
-* []
-
+* [100 - 149] - Reserved for Alarm Class messages to be logged.
+* [150 - 199] - Reserved for Button messages to be logged.
+* [200 - 249] - Reserved for Buzzer messages to be logged.
+* [250 - 299] - Reserved for Component messages to be logged.
+* [300 - 349] - Reserved for ComponentTester messages to be logged.
+* [350 - 399] - Reserved for Laser messages to be logged.
+* [400 - 449] - Reserved for LED messages to be logged.
+* [450 - 499] - Reserved for Photoresistor messages to be logged.
+* [700 - 999] - Currently reserved for David Tille.
 **Notes:**
 * In general, you should never send a message of Type `0`. That really just wouldn't make any sense...
 * `Log Level` is only relevant when a `Log` Type message is sent. In all other cases a `0 - N/A` should be passed.
--- a/docs/development/logging.md
+++ b/docs/development/logging.md
@@ -55,6 +55,12 @@ Message Translation Chart
 |             |                  | 103 - Alarm successful action
 |             |                  | 104 - Alarm failed to arm
 |             |                  | 105 - Alarm successfully armed
+|             |                  | 106 - Alarm successful calibration
+|             |                  | 107 - Alarm has begun arming
+|             |                  | 108 - Alarm has been triggered
+|             |                  | 109 - Alarm calibration is being reset
+|             |                  | 110 - Alarm is being disarmed
+|             |                  | 111 - Alarm is being is being silenced
 |             |                  | 150 - Button object successfully created
 |             |                  | 151 - Button has been pressed
 |             |                  | 200 - Buzzer object has been created
--- a/docs/development/logging.md
+++ b/docs/development/logging.md
@@ -47,11 +47,11 @@ Message Translation Chart
 | 3 - Command | 3 - Debug        | 003 - Execute Command: Silence |
 | 4 - Error   | 4 - Info         | 004 - Execute Command: Calibrate |
 | 5 - Request | 5 - Trace        | 005 - Execute Command: Trigger |
-| | | 006 - Execute Command: Reset Calibration |
-| | | 007 - Execute Command: Test Components |
-| | | 514 - [EXAMPLE] Component Configured on Pin 14 |
-| | | 515 - [EXAMPLE] Component Configured on Pin 15 |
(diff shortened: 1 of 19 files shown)


src/Alarm.cpp
```diff
@@ -13,24 +21,36 @@
...
+*/
 void Alarm::calibrate(){
   determineBasePhotoresistorReading();
   if(_photoR->takeReading() - _threshold < _baseReading){
     alertFailedAction();
     _isCalibrated = false;
   } else {
     alertSuccessfulAction();
-    Logger::Log("Alarm Successful Calibration");
+    Serial.write(30004); // 30004 = Command, N/A, Execute Command: Calibrate
```
Lastrellik: The 30004 command is the command I send from the PI to the Arduino when I want it to calibrate. This isn't the right 5 digit number.

src/Alarm.cpp
```diff
@@ -53,91 +80,138 @@ void Alarm::determineBasePhotoresistorReading(){
...
 }
-
+/**
+  Sets the alarm to be triggered
+  @param sets _isTriggered to be true
+  @param sets _isSilenced to be false
+*/
 void Alarm::trigger(){
-  Logger::Log("Alarm has been triggered!");
+  Serial.write(30005); // 30005 = Command, N/A, Execute Command: Trigger
```
Lastrellik: Again, 000 - 099 are for commands that the PI sends back to the arduino to tell it to do something.

src/Alarm.cpp
```diff
@@ -53,91 +80,138 @@ void Alarm::determineBasePhotoresistorReading(){
...
   if (not this->_isSilenced) _buzzer->soundAlarmLowTone();
   _alarmLED->flash(300);
 }
-
+/**
+  Disarms the alarm and sets the calibration to false to resetCalibration
+  @param _isCalibrated is set to false
+*/
 void Alarm::resetCalibration(){
+  Serial.write(30006); // 30006 = Commmand, N/A, Execute Commmand: Reset Calibration
```
Lastrellik: Same as the above issues.

src/Alarm.cpp
```diff
@@ -53,91 +80,138 @@ void Alarm::determineBasePhotoresistorReading(){
...
+*/
 void Alarm::disarm(){
   this->alertSuccessfulAction();
   _isTriggered = false;
   _isArmed = false;
   _alarmLED->off();
   _isSilenced = false;
   _laser->off();
-  Logger::Log("Alarm has been disarmed");
+  Serial.write(30002); // 30002 = Command, N/A, Execute Command: Disarm
```
Lastrellik: Same as above issues

src/Alarm.cpp
```diff
@@ -53,91 +80,138 @@ void Alarm::determineBasePhotoresistorReading(){
...
 }
+/**
+  Silences the alarm by setting _isSilenced to true
+*/
 void Alarm::silence(){
   _isSilenced = true;
-  Logger::Log("Alarm has been silenced");
+  Serial.write(30003); // 30003 = Command, N/A, Execute Command: Silence
```
Lastrellik: Same as above issues

src/Buzzer.cpp
```diff
@@ -1,45 +1,87 @@
...
+  Uses the int frequency to determine the pitch of the tone being passed to the
+    tone function.
+    This version of the soundTone Function does not use duration with the tone
+    function.
+*/
 void Buzzer::soundTone(int frequency) {
-  if(Properties::DEBUGGING_ACTIVE) Serial.println("Buzzer has started sounding");
+  //Need to take a look at while we're using properies for a Serail.println
+  //does not affected the tone
+  //if(Properties::DEBUGGING_ACTIVE) Serial.println("Buzzer has started sounding");
```
Lastrellik: Yeah we can remove this code. That should be fine.

src/Buzzer.cpp
```diff
@@ -1,45 +1,87 @@
...
 void Buzzer::soundNegativeTone(){
   this->soundTone(400, 300);
 }
-
+/**
+  Stops the tone of the Buzzer sounding
+*/
 void Buzzer::stopTone() {
-  if(Properties::DEBUGGING_ACTIVE) Serial.println("Buzzer has stopped sounding");
+  //if(Properties::DEBUGGING_ACTIVE) Serial.println("Buzzer has stopped sounding");
```
Lastrellik: We can remove this line of code too.

src/Laser.cpp
```diff
@@ -1,26 +1,45 @@
-/*
-  Laser.cpp - Simmple library for the laser module
+/**
+  Calico Home Security System, Laser.h
+  Purpose: Manges the lASER and creats a LASER object
```
Lastrellik: Few Typos here

src/Laser.cpp
```diff
@@ -1,26 +1,45 @@
...
 void Laser::off(){
+  Serial.write(13352); // 13352 = Log, Debug, Laser has been turned off
   digitalWrite(Component::getPin(), LOW);
   _isOn = false;
 }
-
+//Need to think about removing toggle as we don't really use it
 void Laser::toggle(){
   if(_isOn){
     digitalWrite(Component::getPin(), LOW);
```
Lastrellik: We can remove it. I'm sure that's fine.

src/Laser.h
```diff
@@ -1,5 +1,9 @@
-/*
-  Laser.h - Simple class for laser module
+/**
+  Calico Home Security System, Laser.h
+  Purpose: Manges the lASER and creats a LASER object
```
Lastrellik: Manges? Manages? Another Typo :P

Lastrellik (changes requested):
Just a few changes here before we can merge it. Shouldn't take too long.

src/Laser.h
Boomstick105: I'll get right on fixing these issues, I guess I was miss understanding the command function a little bit, shouldn't be to hard to update these real quick.

Boomstick105: Should be all updated now by getting rid of the 30000's and the small typos

docs/development/logging.md
```diff
@@ -47,16 +47,53 @@ Message Translation Chart
...
-
+* [100 - 149] - Reserved for Alarm Class messages to be logged.
+* [150 - 199] - Reserved for Button messages to be logged.
+* [200 - 249] - Reserved for Buzzer messages to be logged.
+* [250 - 299] - Reserved for Component messages to be logged.
+* [300 - 349] - Reserved for ComponentTester messages to be logged.
+* [350 - 399] - Reserved for Laser messages to be logged.
+* [400 - 449] - Reserved for LED messages to be logged.
+* [450 - 499] - Reserved for Photoresistor messages to be logged.
+* [700 - 999] - Currently reserved for David Tille.
```
Lastrellik, line 96: I like this.

src/Component.h
```diff
@@ -1,5 +1,11 @@
-/*
-  Component.h - Component superclass for component classes
+/**
+  Calico Home Security System, Component.h
+  Purpose: Handles creation of a Component object.
+    Which are used to determine the type of compoent Ex. "LED" and where it is
```
Lastrellik: Typo in compoent

src/Laser.cpp
```diff
@@ -1,32 +1,42 @@
...
 }
-
+/**
+  Base contructor
+*/
 Laser::Laser(){
 }
-
+/**
+  Turns the lASER on
```
Lastrellik: Should all letters in LASER be upper case?

Lastrellik (changes requested):
So The changes all look good with the exception of a few typos. I also never loaded the code onto the arduino because it's currently plugged into the PI on the other side of the room. I'll load it up and look at it tonight.

Lastrellik:
@Zeustopher can you load this code and make sure it functions? I think that removing the Laser::toggle() from the cpp but not from the header may cause it to break. I hate C++. My arduino is tied up with the PI at the moment.

Boomstick105:
I forgot to remove that from the laser.h, I edit it a couple of times and then went back I can remove that from the resubmit real quick @Lastrellik

Zeustopher:
@Boomstick105 - There are several build errors when I try and build this branch in PlatformIO. Is it building on properly on your machine?

Boomstick105:
I'm getting errors from the properties and logger class because of a / and
an extra bracket. Just pushed the fix and it built fine.

Lastrellik:
@Boomstick105 we should be able to remove the logger class. We're not building datapackets anymore, so there should be no problem with removing that one.

Zeustopher (approved):
I think there are some additional tweaks we can make to this as we move forward, but this Pull Request is starting to get large and already has quite a few commits to it. I vote that we merge it in as-is right now and work on those additional changes (like removing the Logger class) in another Pull Request.

If anyone else agrees go ahead and merge this pull request.

Lastrellik:
Merged [#55](https://github.com/Lastrellik/Calico-Home-Security/pull/55)

PR #56, opened 2017-03-28
Added IntelliJ project files to .gitignore

Zeustopher:
My Java IDE of choice is [IntelliJ](https://www.jetbrains.com/idea/). Its project files are different from those created by Eclipse, which other team members are using, so I needed to add IntelliJ's project files to the `.gitignore` files so that they don't get committed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,15 @@
 .classpath
 factoryConfiguration.json
 
+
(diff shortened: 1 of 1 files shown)


Lastrellik:
I can't imagine this breaking anything.

PR #59, opened 2017-03-28
Delete 64-byte DataPacket Paradigm Readme

Lastrellik:
This documentation is no longer accurate.


Zeustopher:
@Lastrellik - This same info is in the README inside the circuit diagrams as well. Should that one also be deleted?

https://github.com/Lastrellik/Calico-Home-Security/blob/master/docs/circuit_diagrams/module_PI/README.md

Lastrellik:
@Zeustopher Oh yeah... I guess that's the problem with heavy documentation. It's not always synchronized with what we're actually implemented. When I have a minute I'll fix it.

Lastrellik:
@Zeustopher Alright I deleted the other one too. Should I do the PI Java documentation now too or should I create another branch for that?

Zeustopher:
Lets create another branch for that.

Also - I'm less concerned about the Java Docs at this point than having a README about the module. How to set it up, what the wiring is like from the Arduino to the Pi and the Pi to the touchscreen, how to run the code, which model touchscreen it is you got, etc. That way it stays consistent with our other modules and any person could pick it up, buy the parts, wire it up, and run the code.

Lastrellik:
I'll focus on documentation next. I really want to hand the GUI part off to @Boomstick105 . Meaning minimal involvement.

PR #60, opened 2017-03-30
Wifi Module Refactor

Zeustopher:
This change contains several items related to the Wifi Module:

1. The first is the addition of the `lib_deps` section in the `platformio.ini` file. This uses PlatformIO's [Library management system](http://platformio.org/lib) to pull in the required `WiFiEsp` library.

2. The next main refactor was renaming from `Wifi` to `WifiModule`. I had to do this as the WifiEsp library had a constant named Wifi which caused some compile and runtime issues. The quick-win was just renaming our file instead of working around it.

3. The final refactor was in the wifi code itself. I cleaned out ALOT of the initial code I had done and replaced it with code utilizing the new WiFiEsp library. For right now that code is commented out and I will fully remove it in a future commit.

NOTE - This currently has an issue where everything compiles and runs but the Arduino isn't actually able to communicate with the ESP8266... I suspect it is happening somewhere between the constructors of the `esp8266` SoftwareSerial object and passing that into the ` WiFi.init(&esp8266);` method. It works if I define the SoftwareSerial and do the WiFi.init() in the main class but thats 'icky' and I want to have all of that contained in this class (if possible). I will troubleshoot this more in the future and either go the 'icky' route to get it functioning or identify where the real issue is and correct it.


Zeustopher:
I fixed the above mentioned issue about it not actually functioning. Details are in 73c801b7e5. So it now functions properly!

I will be hooking it up to the alarm in the next commit. I'll probably include it in this Pull Request since it hasn't been reviewed yet and won't be too much more code.

Zeustopher:
Its now hooked up and will send a notification when the alarm is triggered! 🎉

Lastrellik (approved):
Looks good.

Lastrellik:
So I finally ripped apart my PI and all of that to load this up. It looks like it runs just fine. I don't have the WiFi hardware, but the base still seems to run just fine.

PR #61, opened 2017-04-02
Added Future Dev Ideas Readme

Zeustopher:
I had an idea about using MQTT and rearchitecting the project to utilize a publish/subscribe model. This would be a pretty large refactor and one that we probably wouldn't really have time to accomplish by the time this project is due. However, it is a neat idea and one that I thought would be worth documenting and including in the GitHub repository. It would also be a cool idea to include in the whitepaper we are going to be writing for the project.

There are probably other ideas out there like this so I added a new file for any others we want to include and linked to it from the main README.

_(You can just ignore the `MQTT High Level.uxf ` file. Thats the UMLet 'xml' file. No need to sift through all of it an review it. It resulted in the png I exported)_


Lastrellik:
@Zeustopher I mean yeah I'm all for doing this. I feel like this project isn't going to be done when the class is over, at least not for me. I'm game to keep working on it if you are.

Zeustopher:
I had another idea as I was thinking about it that having a web application to arm/disarm/see the status of the alarm would be really cool. Its also very in line with the way IoT is happening right now. Since this Pull Request hadn't been merged yet I just tacked it on to this one.

sithboy:
I agree. I'm definitely going to expand on things beyond this class, probably more simple coding but harder builds. I think it is valid to include these ideas to show that we're thinking about other ways to improve the project.

sithboy (commented):
I didn't even see any typos. Nice and clean.

PR #63, opened 2017-04-02
Pi documentation

Lastrellik:
I created a very basic layout of a README for the Raspberry PI module. 

I could have went through and documented all the steps I took in order to get the image, installation and permissions the way that they are, but I think it might make more since to just take an image of the OS and upload it here so people can download and implement the image directly.

Let me know what you think. I can be more specific or more robust if need be.

### DIFF
--- a/docs/circuit_diagrams/module_PI/README.md
+++ b/docs/circuit_diagrams/module_PI/README.md
@@ -1,7 +1,7 @@
 The Raspberry Pi Module
 ===============
 
-The **Raspberry PI module** integrates a [Raspberry Pi 3](https://www.raspberrypi.org/products/raspberry-pi-3-model-b/) into the circuit to add another layer of functionality. It is initially intended to include the following features:
+The **Raspberry PI module** integrates a [Raspberry Pi 3](https://www.raspberrypi.org/products/raspberry-pi-3-model-b/) into the circuit to add another layer of functionality. It is installed by simply plugging a serial cable from any of the USB ports on the Raspberry Pi into the serial port on the Arduino Uno. It is initially intended to include the following features:
 
 *A 7'' Touchscreen to allow for a GUI
 *A timestamped logging system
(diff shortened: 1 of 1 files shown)


docs/circuit_diagrams/module_PI/README.md
```diff
@@ -0,0 +1,40 @@
...
+  -[This kit](https://www.amazon.com/LoveRPi-Raspberry-Plug-Play-Starter/dp/B01IYBZEV6/ref=sr_1_16?s=pc&ie=UTF8&qid=1491114514&sr=1-16&keywords=raspberry+pi+3) comes with all the hardware you need to get started.
+* [7 Inch Touchscreen](https://www.amazon.com/Raspberry-Pi-7-Touchscreen-Display/dp/B0153R2A9I/ref=sr_1_1?s=electronics&ie=UTF8&qid=1491114627&sr=1-1&keywords=7+inch+touchscreen+raspberry+pi)
+* [Raspberry Pi Touchscreen Case](https://www.amazon.com/Case-Official-Raspberry-Touchscreen-Display/dp/B01HV97F64/ref=pd_sim_147_1?_encoding=UTF8&psc=1&refRID=HCS99YG62WD3F73SRZXA)
+
+
+Setup
+=====
+Since this not much is required for the Raspberry Pi Module to work, it would be wasteful to use the whole Raspbian Operating System to do what we're trying to do. Instead, what we've done is build the OS up from scratch starting with [Raspbian Jesse Lite](https://www.raspberrypi.org/downloads/raspbian/) and then installing the [PIXEL](https://www.raspberrypi.org/blog/introducing-pixel/) shell by following [this guide](https://www.raspberrypi.org/forums/viewtopic.php?f=66&t=133691). We then installed Java and set up a developer and user account and installed VNC and SSH.
```
Zeustopher: Not sure exactly what you are meaning by "Since **this** not much is required". Did you mean it to say "Since not much is required" or possibly "Since there is not much required"?

docs/circuit_diagrams/module_PI/README.md
```diff
@@ -0,0 +1,40 @@
...
+=====
+Since this not much is required for the Raspberry Pi Module to work, it would be wasteful to use the whole Raspbian Operating System to do what we're trying to do. Instead, what we've done is build the OS up from scratch starting with [Raspbian Jesse Lite](https://www.raspberrypi.org/downloads/raspbian/) and then installing the [PIXEL](https://www.raspberrypi.org/blog/introducing-pixel/) shell by following [this guide](https://www.raspberrypi.org/forums/viewtopic.php?f=66&t=133691). We then installed Java and set up a developer and user account and installed VNC and SSH. 
+
+Development
+============
+In order for the Raspberry Pi to communicate with the Arduino through Serial, we needed to add the user to the uucp, dialout, lock and tty groups according to [this guide](https://github.com/Fazecast/jSerialComm/wiki/Installation) from JSerialComm. 
+
+Since there is no Java IDE (nor is there need for one) on this installation of Raspbian, we need to compile and run the Java code from the command line. To simplify this, create a bash script that compiles each of the jar files and specifies the classpath similar to this: 
+
+````
```
Zeustopher: In this particular case it really doesn't have much effect, but when you specify chunks of code like this in Markdown you can actually specify a language at the end of the 3 back-ticks. When you do that it will do some syntax/keyword highlighting. You can specify things like `java`, `javascript`, `python` or in the case `bash` or `shell`.

So something like
```
```java
int foo = 12;
System.out.println("Hello Markdown!");
``
```
would end up looking like 
```java
int foo = 12;
System.out.println("Hello Markdown!");
```

In this particular case I doubt it will color anything except maybe the bash definition comment at the very start of it. So this is more of an FYI comment instead of me asking for you to do anything with it.

docs/circuit_diagrams/module_PI/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+Parts
+-----
+* [Raspberry Pi 3 Model B](https://www.raspberrypi.org/products/raspberry-pi-3-model-b/)  
+  -[This kit](https://www.amazon.com/LoveRPi-Raspberry-Plug-Play-Starter/dp/B01IYBZEV6/ref=sr_1_16?s=pc&ie=UTF8&qid=1491114514&sr=1-16&keywords=raspberry+pi+3) comes with all the hardware you need to get started.
+* [7 Inch Touchscreen](https://www.amazon.com/Raspberry-Pi-7-Touchscreen-Display/dp/B0153R2A9I/ref=sr_1_1?s=electronics&ie=UTF8&qid=1491114627&sr=1-1&keywords=7+inch+touchscreen+raspberry+pi)
+* [Raspberry Pi Touchscreen Case](https://www.amazon.com/Case-Official-Raspberry-Touchscreen-Display/dp/B01HV97F64/ref=pd_sim_147_1?_encoding=UTF8&psc=1&refRID=HCS99YG62WD3F73SRZXA)
+
+
+Setup
```
Zeustopher, line 24: I know that it is as simple as plugging in the USB cable from the Pi to the Arduino to have them communicate, but lets state that somewhere in this section. Or, to be more consistent with our other diagrams, you could add a Wiring section and just make a statement about just hooking the usb ports together. Someone who doesn't know how you've implemented it might be wondering how to get that to work.

There also should be some sort of diagram or statement about how the touchscreen is wired up as well. I don't have a clue what that wires up like to the Pi so I'm as in the dark as someone external would be.

Zeustopher (changes requested):

Lastrellik:
@Zeustopher changes have been made. Again, I appreciate your feedback. Please be as critical as possible; I'm here to learn.

Zeustopher:
👍

PR #69, opened 2017-04-06
Pi junit tests

Lastrellik:
I created a JUnit test suite for the current state of the code on the Raspberry Pi. I plan to implement my current understanding of Test Driven Development in the PI code moving forward.


src/module_PI/Raspberry_PI/JUnitTests/DataPacketTest.java
```diff
@@ -2,33 +2,117 @@
...
+	public void testDataPacketByteArrayFromByteArray() {
+		byte[] packetArray = {1,2,3,4,5};
+		DataPacket packet = new DataPacket(packetArray);
+		assertTrue(Arrays.equals(new byte[] {1,2,3,4,5}, packet.getPacketAsArray()));	
+	}
+	
+	@SuppressWarnings("unused")
+	@Test(expected = IllegalArgumentException.class)
+	public void testDataPacketByteArrayFromByteArray_invalidLength() {
+		byte[] packetArray = {1,2,3,4,5,6};
```
Zeustopher, line 69: You might also want to test sending an empty array. Sometimes empty collections cause different errors to occur than this one that you're catching.

src/module_PI/Raspberry_PI/JUnitTests/SerialInputListenerTest.java
```diff
@@ -4,31 +4,34 @@
...
 	}
 
 	@Test
 	public void testIsDataStreamEmpty() {
-		fail("Not yet implemented"); // TODO
+		SerialInputListener inputListener = new SerialInputListener();
+		assertTrue(inputListener.isDataStreamEmpty());
 	}
 
+/*TODO Have Arduino Respond To Ping Request, then write this test.
```
Zeustopher, line 23: I noticed this comment. I'm not suggesting any changes at this moment but just some ideas to consider when you come back to this comment/test:

For "_Unit Testing_" you don't want to rely on an outside dependency like this. You wouldn't want to have the actual Arduino respond to a ping.

For "_Acceptance Testing_" (_or possibly "Integration Testing" or "System Testing"_) - http://softwaretestingfundamentals.com/acceptance-testing/ you want want to test the 'interaction ' or 'integration' with external dependencies. This helps check to make sure the 'contract' between the two things is still correct. Those tests are generally written in a test file of their own and run separately from Unit Tests. But for Unit Tests you wouldn't want to deal with external dependencies.

To accomplish the test you look like you are wanting to do here there are multiple 'mocking' frameworks out there that will let you accomplish basically what you're wanting to do. In essence it is like making a "fake arduino" that you'd "call" and get a packet response back from. 

One such mocking framework is [Mockito](http://site.mockito.org/) (_another I've heard a bunch about is called [EasyMock](http://easymock.org) though I've never used it myself_). With it you'd end up doing something similar to declaring a mock for the sender and then do a line something like ```when(mockedSerialSender.isPacketAvailable()).thenReturn(true);``` (Obviously I haven't looked in detail at what the actual calls would be but thats just an example). Then you would make the call and move forward with whatever test you would want to accomplish. Reading over the front page of Mockito's website will help make it more clear than the example I showed here.

So thats just some things to chew on and think about in terms of testing. Its quite common to mock things like this in larger projects where there are multiple external dependencies.

Zeustopher (approved):

Zeustopher:
@Lastrellik - I've made a few comments but overall I think this looks good. I'm going to merge it and you can implement the suggestions (if you choose to) on another pull request.

PR #70, opened 2017-04-07
Create PacketMessageTranslator

Lastrellik:
This is my first attempt at TDD and I'm hooked.

I feel like I was able to write this code far more efficiently because I spent less time debugging. First I make the test, then I write the code, then I test it. I immediately see if it's going to work or not rather than wasting time assuming that it's correct only to spend (literally) hours later trying to fix it.

This class translates packets and packet messages into strings using a HashTable. While it does use more RAM, the HashTable approach is a scalable way to convert packets into strings instantly. This will be used as a tool for the logging class which I'm going to write next.

### DIFF
--- a/src/module_PI/Raspberry_PI/JUnitTests/PacketMessageTranslatorTest.java
+++ b/src/module_PI/Raspberry_PI/JUnitTests/PacketMessageTranslatorTest.java
@@ -12,17 +12,17 @@ public class PacketMessageTranslatorTest {
 	@Rule
 	public ExpectedException expectedEx = ExpectedException.none();
 
+	PacketMessageTranslator translator = new PacketMessageTranslator();
+
 	@Test
 	public void testTranslateFromInt() {
 		int executeCommandInt = 1;
-		PacketMessageTranslator translator = new PacketMessageTranslator();
 		assertEquals(translator.translate(executeCommandInt), "Execute Command: Arm");
 	}
 	
 	@Test
 	public void testTranslateFromInt_TooHigh() {
 		int tooLargeDetailInt = 100000;
-		PacketMessageTranslator translator = new PacketMessageTranslator();
 		expectedEx.expectMessage("invalid packet int: 100000");
 		translator.translate(tooLargeDetailInt);
 	}
@@ -30,7 +30,6 @@ public class PacketMessageTranslatorTest {
 	@Test
 	public void testTranslateFromInt_TooLow() {
 		int tooLowDetailInt = 0;
-		PacketMessageTranslator translator = new PacketMessageTranslator();
 		expectedEx.expectMessage("invalid packet int: 0");
 		translator.translate(tooLowDetailInt);
 	}
@@ -38,7 +37,6 @@ public class PacketMessageTranslatorTest {
 	@Test
 	public void testTranslateFromInt_BetweenDetailsAndFullLength() {
 		int intBetweenDetailsAndFullLength = 8000;
-		PacketMessageTranslator translator = new PacketMessageTranslator();
 		expectedEx.expectMessage("invalid packet int: 8000");
 		translator.translate(intBetweenDetailsAndFullLength);
 	}
@@ -46,7 +44,6 @@ public class PacketMessageTranslatorTest {
 	@Test
 	public void testTranslateFromInt_FullPacketLength(){
 		int intBetweenDetailsAndFullLength = 11001;
-		PacketMessageTranslator translator = new PacketMessageTranslator();
 		String expected = "Execute Command: Arm";
 		String actual = translator.translate(intBetweenDetailsAndFullLength);
 		assertEquals(expected, actual);
@@ -56,7 +53,6 @@ public class PacketMessageTranslatorTest {
 	@Test
 	public void testTranslateFromInt_MessageNotFound() {
 		int unmappedInt = 999;
-		PacketMessageTranslator translator = new PacketMessageTranslator();
 		expectedEx.expectMessage("packetDetails 999 not in messageTable");
 		translator.translate(unmappedInt);
 	}
@@ -64,14 +60,12 @@ public class PacketMessageTranslatorTest {
 	@Test
 	public void testTranslateFromString_IntegerString(){
 		String integerString = "001";
-		PacketMessageTranslator translator = new PacketMessageTranslator();
 		assertEquals(translator.translate(integerString), "Execute Command: Arm");
 	}
 	
 	@Test
 	public void testTranslateFromString_IntegerFromStringTooLong(){
 		String tooLongString = "100001";
-		PacketMessageTranslator translator = new PacketMessageTranslator();
 		expectedEx.expectMessage("invalid packet int: 100001");
 		translator.translate(tooLongString);
 	}
@@ -79,14 +73,12 @@ public class PacketMessageTranslatorTest {
 	@Test
 	public void testTranslateFromString_IntegerFromStringTooShort(){
 		String tooShortString = "0";
-		PacketMessageTranslator translator = new PacketMessageTranslator();
 		expectedEx.expectMessage("invalid packet int: 0");
 		translator.translate(tooShortString);
 	}
 	
 	@Test
 	public void testTranslateFromString_NotAnInteger(){
-		PacketMessageTranslator translator = new PacketMessageTranslator();
 		expectedEx.expectMessage("NotAnInteger is not an integer");
 		translator.translate("NotAnInteger");
 	}
@@ -94,21 +86,18 @@ public class PacketMessageTranslatorTest {
 	@Test
 	public void testTranslateFromByteArray(){
 		byte[] validByteArray = {1, 1, 0, 0, 1};
-		PacketMessageTranslator translator = new PacketMessageTranslator();
 		assertEquals(translator.translate(validByteArray), "Execute Command: Arm");
 	}
 	
 	@Test
 	public void testTranslateFromByteArray_MessageDetailsOnly(){
 		byte[] messageDetailsByteArray = {0, 0, 1};
-		PacketMessageTranslator translator = new PacketMessageTranslator();
 		assertEquals(translator.translate(messageDetailsByteArray), "Execute Command: Arm");
 	}
 	
 	@Test
 	public void testTranslateFromByteArray_LongLength(){
 		byte[] longLengthByteArray = {1, 1, 1, 0, 0, 1};
-		PacketMessageTranslator translator = new PacketMessageTranslator();
 		expectedEx.expectMessage("6 is an invalid byte length");
 		translator.translate(longLengthByteArray);
 	}
@@ -116,7 +105,6 @@ public class PacketMessageTranslatorTest {
 	@Test 
 	public void testTranslateFromByteArray_ShortLength(){
 		byte[] shortLengthByteArray = {0, 1};
-		PacketMessageTranslator translator = new PacketMessageTranslator();
 		expectedEx.expectMessage("2 is an invalid byte length");
 		translator.translate(shortLengthByteArray);
 	}
--- a/src/module_PI/Raspberry_PI/JUnitTests/PacketMessageTranslatorTest.java
+++ b/src/module_PI/Raspberry_PI/JUnitTests/PacketMessageTranslatorTest.java
@@ -8,6 +8,7 @@ import org.junit.rules.ExpectedException;
 import module_PI.Raspberry_PI.main.PacketMessageTranslator;
 
 public class PacketMessageTranslatorTest {
+	PacketMessageTranslator translator = new PacketMessageTranslator();
 	
 	@Rule
 	public ExpectedException expectedEx = ExpectedException.none();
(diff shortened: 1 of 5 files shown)


src/module_PI/Raspberry_PI/JUnitTests/PacketMessageTranslatorTest.java
```diff
@@ -0,0 +1,124 @@
...
+	public void testTranslateFromInt() {
+		int executeCommandInt = 1;
+		PacketMessageTranslator translator = new PacketMessageTranslator();
+		assertEquals(translator.translate(executeCommandInt), "Execute Command: Arm");
+	}
+	
+	@Test
+	public void testTranslateFromInt_TooHigh() {
+		int tooLargeDetailInt = 100000;
+		PacketMessageTranslator translator = new PacketMessageTranslator();
```
Zeustopher: I noticed that you are newing up a PacketMessageTranslator in every single test. _If_ you need to do that then you should do it in an `@Before` method (http://junit.sourceforge.net/javadoc/org/junit/Before.html). Then it will do that setup for you automatically before each test is run which will simplify each test and make it easier to read and see what you're testing.

In fact, I don't think you need to create a new one in every test in this case. You should probably just declare it once as a member variable for this class and refer to the same instance in each test.

I'm going to recommend we change this before merging it in to master. It will make the tests easier to read.

_(I noticed the same thing in the prior pull request but somehow my comment on that didn't get saved. I must not have pushed the button on it or something during my review...)_

src/module_PI/Raspberry_PI/main/PacketMessageTranslator.java
```diff
@@ -0,0 +1,110 @@
...
+	private final int NUM_OF_BYTES_IN_PACKET = 5;
+	private final int MAX_DETAILS_ONLY_INT = 999;
+	private final int MIN_FULL_PACKET_INT = 11001;
+	
+	public PacketMessageTranslator(){
+		buildTable();
+	}
+	
+	private void buildTable(){
+		messageTable.put(1, "Execute Command: Arm");
```
Zeustopher, line 19: This seems to be begging to be put into an ENUM. 

The advantage here would be then you could just iterate over the enum in a simple for loop (or Java 8 Stream). 

The advantage in testing could then be to access the enum to get the strings instead of hard coding (if you chose to do that - some school of thought say not to do that and to hard-code them in tests in case anyone changes the string accidentally).

src/module_PI/Raspberry_PI/main/PacketMessageTranslator.java
```diff
@@ -0,0 +1,110 @@
...
+		String packetMessage = messageTable.get(correctedInt);
+		return packetMessage;
+	}
+	
+	private int parsePacketDetailsFromFullPacketInt(int packetAsInt){
+		int packetModThisIsCorrectLength = (int) (Math.pow(10, NUM_OF_BYTES_IN_PACKET_DETAILS));
+		return packetAsInt % packetModThisIsCorrectLength;
+	}
+	
+	private boolean isIntInScope(int packetDetails){
```
Zeustopher: I think this method is named backwards or the logic inside it is all reversed. It seems that you are going to be returning true if it is _not_ in scope which isn't the way the method is named.

src/module_PI/Raspberry_PI/main/PacketMessageTranslator.java
```diff
@@ -0,0 +1,110 @@
...
+		messageTable.put(404, "LED toggling");
+		messageTable.put(450, "Photoresistor object has been successfully created");
+		messageTable.put(451, "Photoresistor taking reading");
+	}
+	
+	public String translate(int packetDetails){
+		if(isIntInScope(packetDetails)) throw new IllegalArgumentException("invalid packet int: " + packetDetails);
+		int correctedInt = parsePacketDetailsFromFullPacketInt(packetDetails);
+		if(!messageTable.containsKey(correctedInt)) throw new IllegalArgumentException("packetDetails " + packetDetails + " not in messageTable");
+		String packetMessage = messageTable.get(correctedInt);
```
Zeustopher: You don't really need to define a new String here just to then return it. You could combine these two lines to just be
```java
return messageTable.get(correctedInt);
```

src/module_PI/Raspberry_PI/main/PacketMessageTranslator.java
```diff
@@ -0,0 +1,110 @@
...
+	}
+	
+	private boolean isIntInScope(int packetDetails){
+		return (packetDetails < 1 
+				|| packetDetails > MAX_PACKET_DETAILS_INT
+				|| packetDetails > MAX_DETAILS_ONLY_INT && packetDetails < MIN_FULL_PACKET_INT);
+	}
+	
+	public String translate(String packetDetailsAsString){
+		if(isStringNumeric(packetDetailsAsString)) return translate(Integer.parseInt(packetDetailsAsString));
```
Zeustopher: An alternative way to do this is just wrap the parseInt call with a try/catch block. Then you don't have to do the extra check up front and assume that it is a numeric. If it isn't, Java will take care of it and throw and exception that you can catch and rethrow or handle.

(You can choose to change this or not. Its up to you.)

src/module_PI/Raspberry_PI/main/PacketMessageTranslator.java
```diff
@@ -0,0 +1,110 @@
...
+	}
+	
+	private boolean packetIsInvalidLength(byte[] packetAsByteArray){
+		return(packetAsByteArray.length != NUM_OF_BYTES_IN_PACKET 
+				&& packetAsByteArray.length != NUM_OF_BYTES_IN_PACKET_DETAILS);
+	}
+	
+	private String packetByteArrayToMessage(byte[] packetAsByteArray){
+		int parsedDetailsIntFromByteArray = parseDetailsIntFromByteArray(packetAsByteArray);
+		return translate(parsedDetailsIntFromByteArray);
```
Zeustopher: Another instance where you could combine these into a single line if you wanted:
```java
return translate(parseDetailsIntFromByteArray(packetAsByteArray));
```

src/module_PI/Raspberry_PI/main/PacketMessageTranslator.java
```diff
@@ -0,0 +1,110 @@
...
+	}
+	
+	private int parseDetailsIntFromByteArray(byte[] packetAsByteArray){
+		StringBuilder parsedIntFromByteArray = new StringBuilder();
+		int packetEndIndex = packetAsByteArray.length - 1;
+		int packetBeginningIndex = packetAsByteArray.length - NUM_OF_BYTES_IN_PACKET_DETAILS;
+		for(int i = packetBeginningIndex; i <= packetEndIndex; i++){
+			parsedIntFromByteArray.append((byte)packetAsByteArray[i]);
+		}
+		return Integer.parseInt(parsedIntFromByteArray.toString());
```
Zeustopher, line 101: I think this call to parseInt could throw and exception that you aren't handling here if it wasn't able to get an int out of it.

Zeustopher (changes requested):

src/module_PI/Raspberry_PI/JUnitTests/PacketMessageTranslatorTest.java
Lastrellik: Thanks for the feedback! What I tried was just declaring one up at the top and having it apply to all the tests, but that didn't work. I'll check out that junit before functionality tonight.

src/module_PI/Raspberry_PI/main/PacketMessageTranslator.java
Lastrellik: What I read in that book was that you run tests with the expected data being hard coded. I figured that if we made it a variable, especially a variable we pull from the class, that it's not really an effective test. 

In regards to the ENUM, I thought about doing a similar data structure, but my hesitation is that the big-O would be N (or I guess N/2, which is still N) instead of 1 for every call, because it would need to loop through each entry to see if it matches our request. This would be slower, but more RAM friendly. I chose the RAM heavy instant one because the Arduino is going to be shooting messages at it pretty quickly. And I was excited to be able to use a HashTable.

src/module_PI/Raspberry_PI/main/PacketMessageTranslator.java
Lastrellik: Oh yep... You're right. Good catch.

src/module_PI/Raspberry_PI/main/PacketMessageTranslator.java
Zeustopher: I wasn't suggesting not using the `messageTable`. I was more suggesting that the value and strings be put into an enum. Then you can make a for-loop which puts each entry in the enum into the `messageTable`. The checks would still go against the `messageTable` but you end up with a much smaller `buildTable()` method. Doing it that way you get some of the best of both worlds.

src/module_PI/Raspberry_PI/main/PacketMessageTranslator.java
Lastrellik: Whatever is the most clean! I can fix this one.

src/module_PI/Raspberry_PI/main/PacketMessageTranslator.java
Lastrellik: I shall do that.

src/module_PI/Raspberry_PI/main/PacketMessageTranslator.java
Lastrellik: Yeah I mean that's right. I'll fix that tonight as well.

src/module_PI/Raspberry_PI/JUnitTests/PacketMessageTranslatorTest.java
Zeustopher: What issues did it give you when you tried to declare it as a member variable?

Lastrellik:
@Zeustopher thanks again for the awesome feedback! I'll make these changes later tonight when I have some time. I just want to get started with the logging system which I again hope will not take all night.

src/module_PI/Raspberry_PI/JUnitTests/PacketMessageTranslatorTest.java
Lastrellik: It's like the methods didn't see that it existed.

src/module_PI/Raspberry_PI/main/PacketMessageTranslator.java
Lastrellik: Ohh I see. I'll look more into this.

Zeustopher: I'll pull down the code a little later during my lunch break and see if I can identify why that might have been. I'll report back. Work on the other corrections and let me see if I can figure out why.

Lastrellik: The tests require the Arduino to be plugged in. I guess I could document that somewhere.

Zeustopher: If tests require an external dependency then they aren't really "Unit Tests". If it relies on an external dependency like that then it isn't a "Unit Test". See my comment on the previous pull request https://github.com/Lastrellik/Calico-Home-Security/pull/69 and specifically look at the comments I made here: https://github.com/Lastrellik/Calico-Home-Security/pull/69/files/825afa15f6f81b6afadf381b11622568dc45f5f8#diff-5421dd2a9b2dcecfc8aafd309c0c8faa

Essentially you should 'mock' the response an Arduino would be sending.

Lastrellik: I'll check it out. Thanks!

Lastrellik: Sooo honestly I want to move on to the logging system for now. These tests work, I think it would be useful for me to read that book first as soon as I finish this book on Git. After we turn it in then maybe I can look at improving these tests.

src/module_PI/Raspberry_PI/main/PacketMessageTranslator.java
Lastrellik: Is it possible to not get an int out of it? It's (indirectly) from a byte array.

src/module_PI/Raspberry_PI/main/PacketMessageTranslator.java
Zeustopher: If you aren't going to make the test not depend on an Arduino being plugged in then adding a comment somewhere relevant would be enough for now.

I agree that getting the logging system done for now is more important that this would be.

Lastrellik:
@Zeustopher I went ahead and made those changes with the exception of creating the enums or accounting for the possibility of Integer.parseInt from throwing a random error, mostly because I don't see how a byte array will parse into anything outside of an integer. Thanks again for the feedback.

src/module_PI/Raspberry_PI/main/PacketMessageTranslator.java
Zeustopher: Valid point.

Zeustopher (approved):

PR #71, opened 2017-04-07
Removed creation of a new PackageMessageTranslator from each test

Zeustopher:
i have removed the creation of a new PackageMessageTranslator from each test. This was just creating a bunch of objects for not much purpose since a single one can be instantiated and used throughout.


Zeustopher:
@Lastrellik - Pull this down and try it on your box. It works fine on mine.

Lastrellik:
@Zeustopher I guess when it wasn't working when I tried it I had another syntax error and had mistaken it for a visibility issue. Thanks again for your help.

PR #74, opened 2017-04-08
Made wifi setup sound a 'waiting' tone on the alarm

Zeustopher:
This change make a 'waiting' sound trigger before the wifi module connects to wireless and a 'success' sound trigger when it successfully connects.

While debugging / working on my wifi module I realized that after the alarm goes through its normal setup procedures it turns control over to the wifi module to do its thing. It was difficult to know when / if it had finally connected to wifi (_which seems to vary anywhere from 5 seconds to almost 30 seconds..._) so I added a way that you can hear and see that it is starting the wifi setup as well as hear and see when it successfully connected.

Unfortunately, once I hand over control to the ESP8266's library I am using things are out of my control. I can't, for example, continue sound the tone every second or continue flashing an led. If we were in a place where we could do multi-threaded operations then I could do that. In theory, if I were to spend a whole bunch of time working on my own library, I could send the command to the ESP8266 chip to start connecting, return control back to the Arduino (since the ESP8266 is a microcontroller itself and can handle its own processing) and then come back periodically to check "Hey - are you connected yet?". But thats not the way these libraries for it are configured to work... So I'm out of luck on that.

So while not great, this at least gives _something_ to let the user know that something is happening....

_(PS - I added spaces above a bunch of the method comments and forgot to take that out before I committed so thats why you see a bunch of empty lines added.)_

### DIFF
--- a/docs/circuit_diagrams/module_WIFI/README.md
+++ b/docs/circuit_diagrams/module_WIFI/README.md
@@ -122,6 +122,7 @@ By default, the ESP8266 communicates at `115200` baud which is too fast for the 
 3. Build then upload `main.cpp`
 4. After the code has uploaded, launch the Terminal - make sure baud rate is set to `115200`
 5. In the Terminal window, type `AT` and press Enter
+  * _If you have issues with this step, try unplugging and plugging back in the Arduino and relaunching the Serial Monitor. Also try relaunching Atom. It appears that it sometimes just gets into a state where it doesn't fully recognize you are sending commands through it._
 6. You should get back a message `OK`
 7. In the Terminal window, type `AT+UART_DEF=9600,8,1,0,0` and press Enter
 8. In the terminal window you should see the cursor moving and either blank or garbled characters will be displayed.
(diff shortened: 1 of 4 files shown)


Lastrellik:
@Zeustopher did you know that you can combine your commits with a git commit -ammend? It just adds it to your previous commit so that each pull request doesn't come with multiple commits. Pretty nifty. Anyway, this looks good. I don't have a WiFi module, so I can't test it outside of basic alarm functionality. The serial is still broken as heck, though.

Zeustopher:
@Lastrellik - Yeah I do that pretty regularly. Only downside is you can't (usually) do it after you've already pushed the branch and created a pull request. Github blocks that. (or at least used to. I haven't tried in a while)

Zeustopher:
Yeah - still the same. You'd have to do a 'force push' (doesn't that just make you think of star wars?!) in order to let you do it. And doing that can end up getting your code, if you aren't careful, into a really weird/bad state.

PR #75, opened 2017-04-09
Create Activity Logger

Lastrellik:
I created this Activity Logger class.

I wrote it to take in the LogLevel and PacketType as arguments as to avoid creating any dependencies. I can test it with the Arduino itself as soon as @Boomstick105 finishes fixing that code.

### DIFF
--- a/src/module_PI/Raspberry_PI/JUnitTests/ActivityLoggerTest.java
+++ b/src/module_PI/Raspberry_PI/JUnitTests/ActivityLoggerTest.java
@@ -1,0 +1,83 @@
+package module_PI.Raspberry_PI.JUnitTests;
+
+import static org.junit.Assert.*;
+
+import org.junit.*;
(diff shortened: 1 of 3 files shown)


Zeustopher (approved):
No feedback from me on this one. Good job. The tests seem really good. 

I couldn't tell from this, and I'm not really familiar with Java until's logger so it may do this automatically, but when you are logging are got Timestamping the log messages? If not then I'd suggest it as a future improvement.

Lastrellik:
@Zeustopher it does automatically do timestamping. I didn't think it was relevant in the testing because I imagine Java's logger to be tried, true and tested.

Zeustopher:
@Lastrellik - cool. Yeah - totally agree. No need for us to test a standard library. I just wasn't familiar with it and wondered whether you had to do anything to enable Timestamping, etc.

PR #76, opened 2017-04-10
Fixed serial output updates

Boomstick105:
We found the bug that caused the serial flood and squashed it.

It's still not perfect, but at least it's not flooding the serial port any longer.


Lastrellik:
That definitely fixed the flooding bug, but there's another bug in communication. I'll have to figure it out.

Zeustopher:
@Lastrellik - Is the issue you found something that we should block this Pull Request going out? Or is it something on the Pi's side that potentially has the issue?

Lastrellik:
@Zeustopher the problem is that Serial.write([5 digit number]) isn't sending the numbers as a byte array, but rather as a bunch of bytes that aren't what we want. Like it's converting integers to bytes or something like that. So it's definitely a problem with the Arduino. We can merge this if we want, but there's still more work to be done.

Zeustopher:
Ok. I'd think we should merge this so the noise is cut out and then can systematically go in and figure out what the other piece is doing.

If you agree then go ahead and merge this.

PR #77, opened 2017-04-10
Build EmailNotifier

Lastrellik:
It is done.
I still need to implement this class with the rest of the project and I still need to make documentation, but this class is designed to use gmail credentials provided in the included txt file in order to send emails with optional attachments to a given email address. 

The tests I wrote were for the portions of the code that I understood, so it's not quite 100% test coverage. If necessary, I can put more effort into expanding the tests, but I feel like for the given variables it would be pretty difficult to break this code; I was quite thorough.

Next I'll create the webcam implementation and pull it all together. We're really close to being in a state where we can turn this project in.


outgoingGmailCreds.txt
```diff
@@ -0,0 +1,8 @@
+//Enter your username for your Gmail account below.
+//On the line below that, enter your Gmail password.
+//In order to send emails through this application, you
+//will need to enable less-secure apps through gmail. 
+//More info can be found here:  https://support.google.com/accounts/answer/6010255?hl=en
+//Antivirus programs like Avast will also block this functionality. 
+testemail@gmail.com
```
Zeustopher: Any particular reason why you went with the text file route instead of using an actual properties file?https://docs.oracle.com/javase/8/docs/api/java/util/Properties.html

I'd suggest we refactor this to use a Prop file instead of going this route. Its really the standard way to handle this type of thing. (_Another fairly widely used method would be to have it pull from Environment Variables or Command Line parameters but I think since you've started with the text file it is a shorter jump to a Properties file_)

In case you haven't worked with Properties files before, here are a few quick tutorials about them:
https://docs.oracle.com/javase/tutorial/essential/environment/properties.html
https://www.tutorialspoint.com/java/java_properties_class.htm

Let me know whether you want to do that refactor on this pull request or in a future one.

src/module_PI/Raspberry_PI/JUnitTests/EmailNotifierTest.java
```diff
@@ -0,0 +1,77 @@
...
+public class EmailNotifierTest {
+	@Rule
+	public ExpectedException expectedException = ExpectedException.none();
+	
+	public static EmailNotifier testEmailNotifier = new EmailNotifier("testAddress@email.com", "testMessage");
+
+	@Test
+	public void testEmailNotifier_EmailAddressAndMessage() {
+		EmailNotifier emailNotifier = new EmailNotifier("test@test.com", "test");
+		assertEquals(emailNotifier.getAttachmentFilePath(), null);
```
Zeustopher: I'm not sure this test is fulfilling its intended purpose. It isn't checking anything related to the Email Address or Message itself. It is only performing a check on the attachment portion of the message.

I can see that you are probably assuming that if you can call the `emailNotifier.getAttachmentFilePath()` then it must have been constructed properly. Thats _alright_, but it isn't very explicit in the test.

There are a few more tests that kindof fall into this same category a little further down.

src/module_PI/Raspberry_PI/JUnitTests/EmailNotifierTest.java
```diff
@@ -0,0 +1,77 @@
...
+	@Test
+	public void testEmailNotifier_AttachmentDoesntExist(){
+		expectedException.expectMessage("File nonexistantfile.txt can't be found");
+		EmailNotifier emailNotifier = new EmailNotifier("test@test.com", "test", "nonexistantfile.txt");	
+	}
+
+	@SuppressWarnings("unused")
+	@Test
+	public void testEmailNotifier_InvalidEmailAddress(){
+		expectedException.expectMessage("notAProperEmailAddress" + " is not a valid email address");
```
Zeustopher: Is there a reason you are concatenating these two string together? Or just a leftover of a code change or something?

src/module_PI/Raspberry_PI/JUnitTests/EmailNotifierTest.java
```diff
@@ -0,0 +1,77 @@
...
+	@Test
+	public void testEmailNotifier_EmailAddressMessageAndAttachment(){
+		EmailNotifier emailNotifier = new EmailNotifier("test@test.com", "test", "README.md");
+		assertEquals(emailNotifier.getAttachmentFilePath(), "README.md");
+	}
+	
+	@SuppressWarnings("unused")
+	@Test
+	public void testEmailNotifier_AttachmentDoesntExist(){
+		expectedException.expectMessage("File nonexistantfile.txt can't be found");
```
Zeustopher, line 37: I'm not as familiar with this particular syntax with expectedException. I don't see an assert as part of this test. Does this do something 'automagic' behind the scenes and check for that exception's message after the below line is called?

src/module_PI/Raspberry_PI/main/EmailNotifier.java
```diff
@@ -0,0 +1,245 @@
...
+import javax.mail.internet.*;
+
+public class EmailNotifier {
+	
+	File gmailCredentialFile = new File("outgoingGmailCreds.txt");	
+	private String gmailUsername;
+	private String gmailPassword;
+	
+  	private String recipientEmailAddress;
+    private String emailSubject = "Calico Home Security";
```
Zeustopher: Small indenting problem here and on the lines below it.

This also has similar issues scattered throughout this file where indentation is wrong/inconsistent.

src/module_PI/Raspberry_PI/main/EmailNotifier.java
```diff
@@ -0,0 +1,245 @@
...
+	public EmailNotifier(String recipientEmailAddress, String message, String pathToAttachment){
+		parseGmailCredentialsFromFile();
+		this.setRecipientEmailAddress(recipientEmailAddress);
+		this.emailMessage = message;
+		this.setAttachmentFilePath(pathToAttachment);
+		buildPropertiesAndCreateSession();
+	}
+	
+	private void parseGmailCredentialsFromFile(){
+		Scanner credentialScanner;
```
Zeustopher: See my comment on the credentials text file. I think we should refactor this to use a Properties file instead of rolling our own solution to get those credentials.

Especially now that I see you already have a `Properties mailSessionProperties`.

src/module_PI/Raspberry_PI/main/EmailNotifier.java
```diff
@@ -0,0 +1,245 @@
...
+		if(attachmentFilePath != null && !new File(attachmentFilePath).exists()) throw new IllegalArgumentException("File " + attachmentFilePath + " can't be found");
+		this.attachmentFilePath = attachmentFilePath;
+	}
+	
+	public String getRecipientEmailAddress(){
+		return this.recipientEmailAddress;
+	}
+	
+	public void setRecipientEmailAddress(String recipientEmailAddress){
+		String hilariousEmailAddressRegex = "(?:(?:\\r\\n)?[ \\t])*(?:(?:(?:[^()<>@,;:\\\\\".\\[\\]"
```
Zeustopher: :scream:

Zeustopher (changes requested):

src/module_PI/Raspberry_PI/JUnitTests/EmailNotifierTest.java
Lastrellik: [Expected Exception](http://junit.org/junit4/javadoc/4.12/org/junit/rules/ExpectedException.html) is basically a unit test that expects a certain exception from the following code. In this case, nonexistantfile.txt doesn't exist, so we expect it to throw an exception that gives that particular message.

Lastrellik: But I could be doing it incorrectly. This is how I interpreted that information.

Lastrellik:
@Zeustopher changes made. Thanks for the feedback!

Lastrellik:
@Zeustopher Myyyyyyyyyyyyyyyy baddddddddddddddddd. It stored my credentials in a properties file in plain text. The password got changed and it was randomly generated, so it doesn't match any of mine or likely anybody else in the world's password. I panicked and tried to change previous commits, but it's late and I didn't know how to do it. If you can fix this asap that'd be great.

src/module_PI/Raspberry_PI/JUnitTests/EmailNotifierTest.java
Zeustopher: Ah - ok. It does look like it does do some 'automagic' in catching that exception and making sure the proper one was thrown.

Lastrellik: Is that bad?

Zeustopher:
@Lastrellik I'll go see how we can get rid of those from the repo's history. You've already taken care of changing the password so we are 'safe' once again. I've only had to do this once before and it was a pain in the rear, but we can get it fixed.

src/module_PI/Raspberry_PI/JUnitTests/EmailNotifierTest.java
Zeustopher: Nope. I just hadn't seen it before and wanted to make sure the tests were doing what we were expecting them to.

Lastrellik:
@Zeustopher it's an easy mistake to make. The password is not a problem, I imagine that the biggest problem would be that that email address could potentially be a target for spam mail. If it takes more than 5 minutes to fix then I don't think it's really worth it.

Lastrellik:
@Zeustopher but I think I fixed everything else.

Zeustopher:
@Lastrellik I did it the 'easy' way instead of the completely 'safe' way. Essentially I just hard reset myself back to the commit before you accidentally added the credentials file and then did a force push of that up to the repo. That basically 'erased' those commits after it (since they didn't have any actual changes). Anyone watching or listening to this repo can probably get at that file with some git sorcery, but at least it isn't out in the open any longer.

You will need to reset (hard) your local branch to make sure you are in the same state as Github.

I would recommend recommitting a blank / placeholder credentials file again in a new commit now.

Lastrellik:
@Zeustopher so we don't actually need to commit that file because it's generated automatically. If you fixed it, we can just merge it if you think it's production quality.

src/module_PI/Raspberry_PI/main/PiApp.java
```diff
@@ -1,37 +1,46 @@
...
 	/**
 	 * @param args
 	 */
 	public static void main(String[] args) {
+		parseArgs(args);
 		SerialInputListener inputListener = new SerialInputListener();
 		Thread listener = new Thread(inputListener, "Listener");
 		listener.start();		
 		SerialOutput output = new SerialOutput();
+    	EmailNotifier notifier = new EmailNotifier("lastrellik@gmail.com", "Today's message");
```
Zeustopher: Was this line intentionally committed? Or was it by accident just for testing?

src/module_PI/Raspberry_PI/main/PiApp.java
```diff
@@ -1,37 +1,46 @@
 package module_PI.Raspberry_PI.main;
 
+import java.io.*;
+import java.util.*;
 
 public class PiApp {
 
 	/**
 	 * @param args
```
Zeustopher: Now that you are accepting parameters you should update this JavaDoc comment to explain what they are.

src/module_PI/Raspberry_PI/main/EmailNotifier.java
```diff
@@ -0,0 +1,239 @@
...
+		this.setRecipientEmailAddress(recipientEmailAddress);
+		this.emailMessage = message;
+		this.setAttachmentFilePath(pathToAttachment);
+		buildPropertiesAndCreateSession();
+	}
+
+	private void loadGmailCredentials() {
+		try {
+			Properties gmailProperties = new Properties();
+			FileInputStream in = new FileInputStream("gmailCredentials");
```
Zeustopher: Since you went with the command line arguments you didn't need to interact with a file at all.

You should just create a `private Properties gmailProperties = new Properties();` in main or here when you are parsing the args. 

Then you would do something like 
```java
gmailProperties.put("gmailUsername", args[0]);
gmailProperties.put("gmailPassword", args[1]);
```
to set the values as you are parsing through the args.

When you needed to get the values out you would use something like `gmailProperties.getProperty("gmailUsername");`.

Using the above you never have to interact with a file at all. See https://www.mkyong.com/java/java-properties-file-examples/ for some examples of working with Properties.

Ultimately, this comes down to this: since you are throwing an exception in main for not passing in command line arguments properly having a file that you later read from is just extra work. If you didn't require parameters if you already have a gmailCredentials file _then_ it would be more reasonable to still be reading from a file like this.

src/module_PI/Raspberry_PI/main/PiApp.java
```diff
@@ -1,37 +1,46 @@
...
-			pause(5000);
-			output.sendResetCalibrationPacket();
-			pause(5000);
-			output.sendCalibratePacket();
-			pause(5000);
+	}
+	
+	private static void parseArgs(String[] args){
+		int numOfArguments = 2;
+		if(args.length != numOfArguments) throw new IllegalArgumentException("Syntax is java PiApp <username> <password>");
```
Zeustopher: See my comment in loadGmailCredentials() in EmailNotifier.java. This is the line I was referring to that makes the writing to and then reading from a file not necessary / just extra work.

Zeustopher (changes requested):

src/module_PI/Raspberry_PI/main/PiApp.java
Lastrellik: Oh, this was left by accident. I think I should put more effort into reading over my code again before I commit and push it. This was after 1 AM last night and I was pretty tired and only half paying attention because my girlfriend was watching a tv show.

src/module_PI/Raspberry_PI/main/PiApp.java
Lastrellik: So, I tried creating a properties file and then calling it from another class and calling the getProperty method and it couldn't find anything. Again, this probably stems from my lack of understanding of the API. I tried googling it and couldn't find anything relevant except for what I tried, which was still pretty heavy.

src/module_PI/Raspberry_PI/main/EmailNotifier.java
Lastrellik: Ohhh I see what you're saying. I'll take another look at this tonight when I have time. I was wondering about the visibility of that method because it seems like just another object with standard visibility.

Lastrellik: Reading that article it seems like you still have to save the credentials to a properties file, like I was doing. Maybe I'm missing something or maybe I'm misunderstanding, but my next fix keeps writing to a properties file. I'm having a hard time understanding the visibility of the properties file. Like, if I create a local variable called gmailProperties in one method and try to call it from another class's method, it can't see it, which is how I understand Java, but doesn't mesh with how I understand your description of Java Properties. If I'm misunderstanding, please clarify. I appreciate your feedback.

Lastrellik:
@Zeustopher changes made once again. See my above comment and keep the feedback coming! You've taught me more real-world stuff than any other class and I appreciate it.

src/module_PI/Raspberry_PI/main/EmailNotifier.java
Zeustopher: Sorry - I should have used a different example. That one showing it writing to a file was more for an example rather than something you have to do. That only confused things... This example is better at illustrating what I was meaning: https://www.tutorialspoint.com/java/java_properties_class.htm 

You would define the Properties object as a Member variable in the class, not inside a specific method. Right now you are defining it inside the `parseGmailCredentials()` method. Thats why you aren't able to access them elsewhere. You only want to do the `put` calls inside one of the methods.

So instead of what you have, you'll want to define it as part of PiApp class:

```java
public class PiApp {

    public static Properties GMAIL_CREDENTIALS = new Properties(); // <<<<<<<<<<<<<<

    /**
     * @param args Gmail Username and Password for sending mail
     */
    public static void main(String[] args) {
        parseArgs(args);
        ......
    }

    private static void parseArgs(String[] args) {
        GMAIL_CREDENTIALS.put("gmailUsername", args[0]);
        GMAIL_CREDENTIALS.put("gmailPassword", args[1]);
    }

......
```
(_Note the change in name since it is a static member variable and we want to follow standard naming conventions._)

Now anywhere else outside this class you should be able to access it via something similar to `PiApp.GMAIL_CREDENTIALS.getProperty(...);`

Doing it this way you'll never need to access any file. Does that make more sense?

Lastrellik: Okay, yeah it does.
In my defense! I am familiar with how java visibility works, but I thought you were saying that the properties file belong to Java or the Java system, making them outside of the scope of regular java code. I now see the error of my ways.
I'll fix it tonight.

Lastrellik:
@Zeustopher after all this madness, we're ready to take another look at it. Lol 20 commits later. This is how you learn, right?

Zeustopher:
Oops - looks like you didn't do a hard reset like I said you would need to after I fixed the previous issue where you committed the credentials file accidentally. You essentially re-did all that I un-did because you didn't reset so on your next push you put it all back in as it was before the fix. The credentials file now shows up again in the git history. 

Since there have been later changes in the branch I can't do the 'easy' route the same way this time. I think we can fix it, but it will be more difficult. Lets look at it together before/after class tonight and decide whether we want to try and fix it or not.

Lastrellik:
@Zeustopher lol so I'm not really too worried about it. I changed the password and it's not the same as any of my other passwords, and I don't mind getting spam mail to that folder. Let's just accept it.

src/module_PI/Raspberry_PI/main/PiApp.java
```diff
@@ -1,38 +1,27 @@
...
-			pause(5000);
-			output.sendResetCalibrationPacket();
-			pause(5000);
-			output.sendCalibratePacket();
-			pause(5000);
-		}
+	}
+	
+	private static void parseArgs(String[] args){
+		GMAIL_CREDENTIALS.put("gmailUsername" , args[0]);
```
Zeustopher, line 25: Sorry I didn't catch this before - this method now has the possibility of throwing an exception if the person running the app doesn't provide the username and password. In a previous commit you had a check `if(args.length != numOfArguments) throw new IllegalArgumentException("Syntax is java PiApp <username> <password>");` to catch this issues. Either this check or something similar should probably be put back in.

Zeustopher (changes requested):

Lastrellik:
@Zeustopher keep the feedback coming! I made that change.

Zeustopher (approved):

PR #78, opened 2017-04-12
Repair Serial.write byte array bug

Lastrellik:
I was able to repair the byte array bug we had from earlier and correctly parse it out in the Java code. 

It works beautifully. 

I had to switch it from an integer to a string, but since all of the strings are the same size I don't think we'll have the same fragmentation problem we had before because each string can fit in the void made by the removal of any other string in memory.

A bug has been uncovered: When the alarm is armed, it floods the serial port with "Photoresistor taking reading" logs. I can't imagine this being a hard thing to fix.

I also want to change how I handle the Activity Logger. Before I didn't want to create a dependency on an external class, so I was passing 3 arguments to the log method in the ActivityLogger. Now I think I'll change it so that it takes either a String or a DataPacket and parses out all the information there. This can be fixed easily.


Zeustopher:
If we're dealing with Strings again can we / should we also just go back to using `Serial.print` instead of `Serial.write`? Or would that cause issues on the PI's side?

I'm not in a place where I can test it but I'd like to see if we can now see the output in the Serial Monitor inside PlatformIO once again.

Lastrellik:
I'm sure we can. I imagine that it will have the same result either way,
but I'm not sure. It wouldn't be difficult to make that change.

Zeustopher (approved):

PR #79, opened 2017-04-14
Convert to maven project

Lastrellik:
Soooo I decided that I would learn what Maven is and how to use it, mostly because the webcam library depended on it. Anyway, this is my first time dealing with Maven at all. I decided to split my work into two commits, this one and the webcam one, just in case this whole maven thing breaks everything forever. The webcam branch depends on this one working, though. 

Please give feedback. Thanks


Zeustopher:
@Lastrellik - please see https://github.com/Lastrellik/Calico-Home-Security/pull/81 for a few suggested additions to the Maven pom before we merge this into master.

Zeustopher:
Also, just as an FYI, if you ever _really_ don't want to use Maven (I don't know why you wouldn't really...) then you can always go and download the actual jar file from the Maven Central Repository and just import it into your `lib` directory like you did with the others before. You can search the repository at https://search.maven.org/#search%7Cga%7C1%7Ccom.github.sarxos. I like having Maven in the project though so I don't want us to backtrack on this pull request. But I thought it was worth mentioning for future reference in case it is helpful.

Zeustopher (approved):

PR #80, opened 2017-04-14
Build Webcam Functionality

Lastrellik:
Once I figured out the whole Maven thing this wasn't so bad.

This class adds the functionality of being able to take and save pictures using the attached webcam.
Due to the nature of the driver that's hard coded, the webcam will only work on Linux operating systems. 

The library is pretty neat, too. In the future it might be a good idea to look at maybe having it show a live stream on the touchscreen while waiting to be disarmed. It does things like facial recognition and drawing the lines on the face and stuff. It's pretty freaking sweet.

### DIFF
--- a/src/module_PI/Raspberry_PI/main/Camera.java
+++ b/src/module_PI/Raspberry_PI/main/Camera.java
@@ -1,0 +1,64 @@
+package module_PI.Raspberry_PI.main;
+
+import java.awt.Dimension;
+import java.io.File;
+import java.io.IOException;
+
+import javax.imageio.ImageIO;
+
+import com.github.sarxos.webcam.*;
+import com.github.sarxos.webcam.ds.v4l4j.V4l4jDriver;
+
+public class Camera {
+	
+	private String filePath;
+	private Webcam webcam;
+	private Dimension[] imageResolutions = new Dimension[] {
+			WebcamResolution.PAL.getSize(),
+			WebcamResolution.HD720.getSize(),
+			new Dimension(2000, 1000),
+			new Dimension(1000, 500),
+};
+	public Camera(){
+		this("./");
+	}
+	
+	public Camera(String filePath){
+		setFilePath(filePath);
+		setWebcamDriver();
+		buildWebcam();
+	}
+	
+	private void setWebcamDriver(){
+	    Webcam.setDriver(new V4l4jDriver());
+	}
+	
+	private void buildWebcam(){
+		webcam = Webcam.getDefault();
+		Webcam webcam = Webcam.getDefault();
+		webcam.setCustomViewSizes(imageResolutions);
+		webcam.setViewSize(WebcamResolution.HD720.getSize());
+	}
+	
+	public void takePicture(String pictureFileName){
+		String pictureFileNameWithExtension = pictureFileName;
+		boolean hasFileExtension = pictureFileName.matches("^[\\w,\\s-]+\\.[A-Za-z]{3}$");
+		if(!hasFileExtension) pictureFileNameWithExtension = pictureFileName + ".jpg";
+		try {
+			webcam.open();
+			ImageIO.write(webcam.getImage(), "JPG", new File(this.filePath + "/" + pictureFileNameWithExtension));
+			webcam.close();
+		} catch (IOException e) {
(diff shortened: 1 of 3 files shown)


Lastrellik:
@Zeustopher did you look at this one?

src/module_PI/Raspberry_PI/main/Camera.java
```diff
@@ -0,0 +1,64 @@
...
+public class Camera {
+	
+	private String filePath;
+	private Webcam webcam;
+	private Dimension[] imageResolutions = new Dimension[] {
+			WebcamResolution.PAL.getSize(),
+			WebcamResolution.HD720.getSize(),
+			new Dimension(2000, 1000),
+			new Dimension(1000, 500),
+};
```
Zeustopher: Tiny indenting issue.

src/module_PI/Raspberry_PI/main/Camera.java
```diff
@@ -0,0 +1,64 @@
...
+	}
+	
+	public Camera(String filePath){
+		setFilePath(filePath);
+		setWebcamDriver();
+		buildWebcam();
+	}
+	
+	private void setWebcamDriver(){
+	    Webcam.setDriver(new V4l4jDriver());
```
Zeustopher: Indenting issue.

src/module_PI/Raspberry_PI/main/Camera.java
```diff
@@ -0,0 +1,64 @@
...
+	}
+	
+	private void buildWebcam(){
+		webcam = Webcam.getDefault();
+		Webcam webcam = Webcam.getDefault();
```
Zeustopher: You are constructing another local webcam object here. This makes it difficult to know whether your `setCustomViewSizes` and `setViewSize` are being set on the `webcam` that is a part of the _class_ or whether it is being set on the `webcam` local to this _method_ (which will fall out of scope and be eligible for garbage collection after you leave this method).

I think the intention here was to set these things on the _class_ `webcam` so I would say that this line needs to be deleted.

src/module_PI/Raspberry_PI/main/Camera.java
```diff
@@ -0,0 +1,64 @@
...
+import java.awt.Dimension;
+import java.io.File;
+import java.io.IOException;
+
+import javax.imageio.ImageIO;
+
+import com.github.sarxos.webcam.*;
+import com.github.sarxos.webcam.ds.v4l4j.V4l4jDriver;
+
+public class Camera {
```
Zeustopher, line 12: I'm not sure where to add this comment so I'll just put it here. There needs to be a little documentation on this class as well as in the readme for the specifics on the camera. It should include things like how to hook the camera up, the fact that you've got the driver hardcoded so it will only run on Linux operating systems, what the camera is actually going to be used for in the functionality of the alarm, etc.

Another thought is that I don't know that we want to require a user with the Raspberry Pi module to have a Webcam. If they don't have one, how are we going to handle that? In the main PiApp just don't initialize one? Take that scenario into consideration on your next Pull Request for this camera stuff.

Neither of these need to be added to this Pull Request, unless you want them to, but should be added in the next Pull Request you make.

Zeustopher (changes requested):

Zeustopher:
@Lastrellik - Yes I did but it looks like I never finished and pressed the button to submit the feedback yesterday when I was looking at it at work during my lunch break. Sorry about that. I've redone the review and I think I've added all the comments I had found yesterday.

Lastrellik:
@Zeustopher thanks for the feedback. I found one small error in the pom file but I'll fix that in another commit. Basically we're excluding all the .java files instead of the .cpp files, I think?

Anyway, I made those changes, but I haven't made the documentation yet, mostly because I'm not 100% sure how it's all going to work yet. My thought process is that I'll get it all integrated and then make the documentation, but that could just be an excuse to not write documentation. It'll get done :)

Zeustopher:
@Lastrellik - I thought you had it that way intentionally in the pom. What you can do is run it and then open up the output directory and see what files actually get compiled and see if it looks right. 

As long as it gets documented that's fine. I've just seen too many things in other projects go undocumented and I don't want to let us fall into that trap.

PR #83, opened 2017-04-16
Integrate pi components

Lastrellik:
I know this is another big one, and there's still some work to be done, but we're on a bit of a time crunch. 

This is what I was able to whip together in order to get the different components integrated and working together. I decided to create a whole other thread that processes the packets after they're created to avoid overflows from the Arduino's puny serial buffer. This PacketProcessor class just checks to see if packets have been created by the SerialInputListener and processes them accordingly (logging, taking pictures, emailing pictures, etc). I also had to fight with Maven quite a bit because when we switched over I put in the wrong dependency for the java mail system, which didn't cause syntax errors, but the emails just wouldn't send anymore. I ended up troubleshooting that issue for maybe 4 hours yesterday. 

I also added logging functionality to the classes in the PI itself. 
As far as issues go, and I think we should document these:

- The code doesn't run from the command line properly. I was able to get it to run by executing the following command:  `mvn exec:java -Dexec.mainClass="module_PI.Raspberry_PI.main.PiApp" -Dexec.args="calicoautomation@gmail.com <the calico password I had to change again because I freaking committed it again in the pom file earlier> <all recepient email addresses separated by a space>"`
- I put a ten second timer after the picture got taken to account for saving that picture, but I think that's actually unnecessary.
- I need to write a test for the PacketProcessor. I didn't know how to do it because we, again, depend on the arduino being plugged in for it to work. I'll fix this when we have a little more time.
- The arduino still floods the serial port when it's armed with the "Photoresistor taking reading" messages. That will need to be fixed.
- There is some hardcoded stuff throughout the project that I think needs to be a little less.... hardcoded. Like the email subject, etc.
- I know it's possible to send an email to multiple people at the same time instead of each recipient one at a time, but I got the emails to send one recipient at a time for now. It's slower, and it gives the intruder more time to smash the equipment before the picture gets sent.
- The pictures I have hardcoded as a JPG, but I don't actually like the quality that comes out of them. There's that notorius jpg effect around the edges of items in the picture. I think we should change it to a png
- The driver we're using for the webcam on the PI keeps giving us this error message: 
````[ libvideo.c:68 ] Using libvideo version 0.9-UNKNOWN
[ v4l2-query.c:49 ] Error looking up V4L2 format 34363248. Please submit
a bug report on the v4l4j mailing list.[ v4l2-query.c:484 ] libvideo has skipped an unsupported image format:
[ v4l2-query.c:485 ] H.264 (0x34363248)
[ v4l2-query.c:486 ] Please let the author know about this error.
[ v4l2-query.c:487 ] See the ISSUES section in the libvideo README file.
205 [module_PI.Raspberry_PI.main.PiApp.main()] INFO com.github.sarxos.webcam.WebcamDiscoveryService - Discovery will not run - driver V4l4jDriver does not support this feature
````
but the camera is taking pictures and stuff just fine. This probably needs to be fixed, but who knows.
- Every time I unplug and plug back in the webcam, it doesn't preserve the permissions. The owner of the camera gets switched to root, so my developer user can't access the camera until I manually go in and `chown developer:developer video0` so I'll need to find a way for that to happen automatically, like a bash script or something.

Sooooo there's still a lot that needs to be done and little time to do it, but I think we can make due for the time being. I also have other homework that I need to do. For now I think my priority is fixing a few things on the arduino and getting the GUI to work. We can fix these small bugs later.

### CODE REVIEW

File: src/module_PI/Raspberry_PI/main/DataPacket.java [hunk 1]
```diff
@@ -60,6 +65,12 @@ private void parseLogLevelFromByteArray(){
 		packetLogLevel = PacketLogLevel.getLogLevelFromByte(packetCodeByteArray[LOG_LEVEL_INDEX]);
 	}
 	
+	private void parseMessage(){
+		int packetIntModThisIsMessageInt = (int) Math.pow(10, NUM_OF_MESSAGE_BYTES_IN_PACKET);
```
Zeustopher: This name was confusing to try and figure out what it was. I had to read over it a few times and look at the code below to figure it out. Possibly breaking from convention here and naming it something like `packetInt_modThisInt_isMessageInt` might make it a little more clear to an outsider looking at this code. 

Its a fairly complex idea you are trying to capture in the name of this variable name. Maybe take a step back and name what it _is_ instead of _what it will be used for_.

src/module_PI/Raspberry_PI/JUnitTests/EmailNotifierTest.java
```diff
@@ -80,4 +80,16 @@ public void testSetRecipientEmailAddress_InvalidEmailAddress(){
 		expectedException.expectMessage("notanemailaddress is not a valid email address");
 		testEmailNotifier.setRecipientEmailAddress("notanemailaddress");
 	}
+	
+	@Test
+	public void testRemoveAttachment(){
+		testEmailNotifier.removeAttachment();
```
Zeustopher: If the default state of an emailNotifier is to have no attachment then you should add an attachment and then remove it in order to prove your method is working. Otherwise, your method might be completely broken but you'd never know it because your emailNotifier is already sitting in the correct state.

pom.xml
```diff
@@ -7,13 +7,12 @@
 	<name>Calico Home Security</name>
 	<build>
 		<sourceDirectory>src/module_PI/Raspberry_PI</sourceDirectory>
-		<!--<testSourceDirectory>src/module_PI/Raspberry_PI/JUnitTests</testSourceDirectory>--> <!-- TODO: Adjust tests so they don't rely on an Arduino being plugged in (should be mocked) and running then uncomment this line so that all unit tests are run at build time -->
+		<testSourceDirectory>src/module_PI/Raspberry_PI/JUnitTests</testSourceDirectory> <!-- TODO: Adjust tests so they don't rely on an Arduino being plugged in
```
Zeustopher, line 10: Were the tests changed so that the Arduino being plugged in isn't necessary? If not then these will fail on someone's machine who doesn't have one plugged in. The dependency on the Arduino should be mocked so it is no longer a dependency.

src/module_PI/Raspberry_PI/main/EmailNotifier.java
```diff
@@ -37,11 +38,13 @@ public EmailNotifier(String recipientEmailAddress, String message, String pathTo
 	}
 
 	private void loadGmailCredentials() {
-			gmailUsername = PiApp.GMAIL_CREDENTIALS.getProperty("gmailUsername");
-			gmailPassword = PiApp.GMAIL_CREDENTIALS.getProperty("gmailPassword");
+		PiApp.LogToFile("EmailNotifier credentials have been loaded");
```
Zeustopher: Shouldn't this be logged after the properties are actually set? Its conceivable that something goes wrong in loading these properties so you don't want to 'tell the future' with your log statements. You could change the way it is worded and have it say `EmailNotifier credentials are being loaded` so you are saying what is coming next instead of assuming that it will work properly.

Zeustopher (changes requested):

src/module_PI/Raspberry_PI/main/DataPacket.java
Lastrellik: Lol I spent a good 5 minutes thinking about what to call that stupid thing. I could just hard code packetInt % 1000, but I see that as being a source of frustration if we ever decide we need an extra digit or something.

pom.xml
Lastrellik: No, I just ran the tests with the Arduino plugged in. I should read about that mocked arduino. I'm just trying to prioritize

src/module_PI/Raspberry_PI/main/EmailNotifier.java
Lastrellik: Oh yeah, good point.

Zeustopher:
@Lastrellik - None of the things I mentioned on my first (relatively quick) pass through of this are super high priority. I think that your statement about starting to be crunched for time is accurate and we may decide to come back and address them later. There are some improvements, etc. that I would suggest if we wanted to get super picky, but I don't think we have the time since we need to switch gears to get a working gui on your touchscreen before middle of the week when we mark a `1.0` release.

If its ok with you, @sithboy & @Boomstick105, I think we are alright to go ahead and merge this.

Lastrellik:
@Zeustopher I made your requested changes with the exception of the virtual arduino. Can we just merge it?

Zeustopher:
For the sake of hitting our deadline, we're merging this with some known areas where we can improve it.

PR #84, opened 2017-04-17
Improve arduino functionality

Lastrellik:
This pull request does a couple things:

- Implements the "tripped" functionality in the project that gives the potential intruder 30 seconds to disarm the alarm (or smash the pi)
- Removes more code that floods the serial port
- fixes incorrect packet ints in Arduino's CommandListener 

I think that's it. It works on the PI and Arduino, but I haven't tested it with any GUI. There may be more bugs to fix in the future.


Lastrellik:
This is @Zeustopher from @Lastrellik's account. We did a code review together and this is approved.

PR #86, opened 2017-04-18
Pi gui

Lastrellik:
I don't know if @Boomstick105 has started to make changes on this branch, but this code allows us to finally have a working GUI with the rest of the base module. It still pops up message boxes when we send test commands and it disarms even when we don't type in a password, but it functions. It logs, sends emails, takes pictures, turns on and off the laser and calibrates everything. It's very very very close to a minimum viable product. 

I really got to get to documenting this stuff, though. There are all kinds of maven dependencies and setup instructions and things to look out for. I also need to hose the PI image and clean it up a little. Also, today when I armed the alarm before I left it ended up triggering like a minute after I left and I think it's because the mirror moved slightly when I closed my front door. I checked my emails like an hour and a half later and saw that I got a picture of my empty, dark apartment. I had to call my girlfriend to go check it out and I guess it was beeping in my house all morning.......................... So we need to account for that kind of stuff. But she was able to figure out how to disarm it, which was pretty sweet.


PI_Log.txt
```diff
@@ -0,0 +1,4 @@
+2017-04-17 23:06:25 Pi SerialOutputListener created
```
Zeustopher: This file shouldn't have been checked in. We should add it to the .gitignore file.

PI_Log.txt.1
```diff
@@ -0,0 +1,2 @@
+2017-04-17 23:49:00 Pi SerialOutputListener created
```
Zeustopher: This file shouldn't have been checked in. We should add it to the .gitignore file. 

It won't let me comment on it directly, but there is also an image which was checked in that shouldn't be. We need to add the entire `PI_PICS` directory to the .gitignore file as well.

Zeustopher (changes requested):
Oh boy... how do I put this one... Without trying to sound harsh, this branch/pull request is a mess... 

I'm seeing junk files like `AddUserPanel.java.old` and `AddUserPanel.java~Updated upstream - Copy` checked in which just seem to be files sitting out there and are duplicates of the actual file. 

Then there are files like `RemoveUserPanel.java.old ` and `RemoveUserPanel.java.current` which I don't know whether it is _actually_ current but I have a hard time thinking it is because a java file isn't named with anything other than a .java file extension.

I also see several files which were checked in unintentionally.

This branch is so long-lived and has had master merged into it so many times (and in fact it actually merged into itself once...) that I can't even try to navigate my way back through it to see how we could squish some of the commits / start a new branch off the current master and cherry pick the actual code change commits.

I just can't give approval for this Pull Request in its given state...

Boomstick105:
So the java .old files are messed up because of me trying to stash, accept,
keep my branch current. I was unaware that they were up there.

So anything .old is a labeling trick, just to keep them around,  I was
having issues with my files being wiped when checking the branch out. If we
need to make a new branch we can,  I was having troubles pulling and
pushing because of those log files. If we need to scrap the gui and start
it over we can. I'll get right on it.

Zeustopher:
@Boomstick105 - We don't want to scrap the work. We just need to clean it up a little bit.  Depending on the route we take it may get a little hairy in git-land.

@Lastrellik - As the initiator of the Pull Request, do you want to clean things up? Or would you like me to take a stab at it?

I think we've got two routes: 1) Just delete the junk files, fix the .gitignore so it doesn't happen again, and then continue with the current Pull Request. 2) We go a more drastic route and try and clean up not only the code but the git history as well via some cherry-picking on a new branch. 

Route 1 probably is almost for sure the simplest and fastest to accomplish. Route 2 is arguably the best, but we're almost at our deadline and I'm not sure we want to sacrifice the time to do then when we could be focusing on documentation and the whitepaper and the presentation...

As much as it pains me, my vote is for Route 1.

Boomstick105:
The main code isn't too much of an issue and it works. I can take the code,
even if it's raw text,  paste it in and submit it a new clean branch just
with the code and as long as the work Jason did in the other classes isn't
too hard to replicate,  we can.  That should clean it up and start it back
over like it's new if my logic is sound.

Lastrellik:
@Zeustopher thanks for the feedback. I was a little hasty to initiate the pull request. I was eager to get a minimum viable product out and told @Boomstick105 to just push whatever he had, then I didn't even take a second to read through the code outside of the bare minimum I needed to get it to function. I thought that if I started trying to fix things that I would go down another 8 hour rabbit hole and I just really don't have the time. I have a Math test on Thursday and I'm underprepared, and I need to do well on this test or I might not get an A in the class. I spent the whole weekend on this project and my other classes homework. @Zeustopher if you can help or literally anyone else, that would be super helpful. It just sucks because all these components depend on each other and nobody else has a PI.... 

This weekend I'll have time to dive in more. The pics and the text files are carelessness on my part. I'm just anxious to get this out.

Zeustopher:
I cleaned up those files and I'm going through the code review. I am likely going to be a little lighter on my comments since we are under the gun.

src/module_TOUCHSCREEN/StartupLogoScreen.java
```diff
@@ -0,0 +1,5 @@
+package module_TOUCHSCREEN;
+
+public class StartupLogoScreen {
+
```
Zeustopher, line 4: Empty?

src/module_TOUCHSCREEN/TouchscreenApp.java
```diff
@@ -1,9 +1,26 @@
...
+ *
+ */
+public class TouchscreenApp {
+	/**
+	 * Main class creates JFrame to house CalicoArmGUI for testing
+	 * @param args
+	 */
+	public static void main(String[] args){
+		
+		JFrame frame = new JFrame("FrameDemo");
```
Zeustopher, line 19: This is all duplicated in the PiApp. Is it needed here?

Zeustopher (approved):

Zeustopher:
I've added only a minimal amount of comments that should be looked at in the not too distant future. I would say we need to get much more focused on the documentation though so these fixes can come later after Release 1.0.

src/module_TOUCHSCREEN/TouchscreenApp.java
Lastrellik: No, I duplicated it. We can remove it from here.

PR #90, opened 2017-04-22
Gui improvement

Lastrellik:
So I went ahead and rebuilt what we had for a GUI. There are several changes in this merge, but it seems to be working flawlessly.

Here's a video I made that shows a little bit of what I did:
https://www.youtube.com/watch?v=_hQCjnNGu70

I know this is last minute, but let me know what you think!

Thanks.


Zeustopher (approved):

Zeustopher:
Great walkthrough video. Looks really great. Can't wait to see it in person.

For the sake of time, I only really would call my code review a "skimming" rather than an in depth review. Nothing blaring jumped out at me.

PR #92, opened 2017-04-26
Added Laser Aiming Mode & Moved Trip Time to Properties

Zeustopher:
* I moved the amount of time between Tripped and Triggered to be in the Properties files. This way it is easier to configure with all of the other application settings.

* It was very useful while setting up our demonstration for the Raspberry Pi to be able to turn on the laser and just leave it on. I implemented a similar mechanism for the base module so you can flip a flag in the Properties file and it will toggle the laser on and leave it on until you press the button. This will help in aiming.


Lastrellik:
Loaded it up and tested it out. Everything still works. I like what you did with the base module and the initial laser.

PR #93, opened 2017-04-26
Improved Wifi Connect Flow

Zeustopher:
I wasn't handling the case where I couldn't connect to wireless. This change allows a configurable number of retry attempts before returning, alerting the user that something is wrong, then continuing with the rest of the alarm functionality.


Lastrellik:
I don't have the WiFi module, but it looks good!

PR #102, opened 2018-10-04
Added GitHub Profile Links to Contributor Names

Zeustopher:
I added our GitHub Profile Links to each of our names in the Contributor Names list (because why not, right?)

(I was glancing at my Starred Repos and this was on there and I got a little sentimental. I'll have to go dig out my project and fire it up sometime after this semester is over.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,10 +60,10 @@ About the Project
 Authors
 -------
 
-* Christopher Nash
-* Tyler Jacobs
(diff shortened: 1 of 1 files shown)


Lastrellik:
Nice! I was thinking about this project today and yesterday too because it would be beneficial in my current situation. That's cool of you to do this.

